PR #3, opened 2016-04-08
reorganized as a Julia package

tanmaykm:


ViralBShah:
Should we enable travis too?

tanmaykm:
Yes, we could.

ViralBShah:
Enabled tests in .travis.yml and turned on CI.

PR #5, opened 2016-05-06
faster tuple comparison and searchsorted

tanmaykm:


JeffBezanson:
Let's improve the definitions in Base instead.

JeffBezanson:
See ongoing discussion in https://github.com/JuliaLang/julia/issues/15695 for some ideas on how to address tuple performance.

ViralBShah:
Let's reopen if necessary.

PR #7, opened 2016-05-23
make NDSparse an AbstractArray

tanmaykm:
~~Should `NDSparse` be an `AbstractArray`?~~

This makes `NDSparse` behave like a basic `AbstractArray`

As an `AbstractArray` it could be used at other places that work with generic abstract arrays.
My immediate goal is to use it with `AxisArray`. Just having `getindex`/`setindex`/`size`/`length` would be enough for that. It can be extended to other methods in future if useful


tanmaykm:
This now has the changes for NDSparse to be a basic AbstractArray.
The constructor now takes size as a parameter. E.g.: `A = NDSparse((10,10), [1:10;], [1:10;], [1:10;])`.

I am able to use this with AxisArrays:

````julia
julia> A = NDSparse(spdiagm(1:5))
NDSparse{Int64,2,Tuple{Int64,Int64}}:
 (1,1) => 1
 (2,2) => 2
 (3,3) => 3
 (4,4) => 4
 (5,5) => 5

julia> X = AxisArray(A, Axis{:c1}([:one, :two, :three, :four, :five]), Axis{:c2}(-5:1:-1))
2-dimensional AxisArray{Int64,2,...} with axes:
    :c1, [:one,:two,:three,:four,:five]
    :c2, -5:1:-1
And data, a NDSparse{Int64,2,Tuple{Int64,Int64}}:
 (1,1) => 1
 (2,2) => 2
 (3,3) => 3
 (4,4) => 4
 (5,5) => 5

julia> X[:one, -4 .. -2]
2-dimensional AxisArray{Int64,2,...} with axes:
    :c1, [:one]
    :c2, -4:1:-2
And data, a 1x3 SubArray{Int64,2,NDSparseData.NDSparse{Int64,2,Tuple{Int64,Int64},Tuple{Array{Int64,1},Array{Int64,1}},Array{Int64,1}},Tuple{UnitRange{Int64},UnitRange{Int64}},0}:
 0  0  0

julia> X[:three, :]
2-dimensional AxisArray{Int64,2,...} with axes:
    :c1, [:three]
    :c2, -5:1:-1
And data, a 1x5 SubArray{Int64,2,NDSparseData.NDSparse{Int64,2,Tuple{Int64,Int64},Tuple{Array{Int64,1},Array{Int64,1}},Array{Int64,1}},Tuple{UnitRange{Int64},Colon},0}:
 0  0  3  0  0

julia> X[[:one,:two], :]
2-dimensional AxisArray{Int64,2,...} with axes:
    :c1, [:one,:two]
    :c2, -5:1:-1
And data, a 2x5 SubArray{Int64,2,NDSparseData.NDSparse{Int64,2,Tuple{Int64,Int64},Tuple{Array{Int64,1},Array{Int64,1}},Array{Int64,1}},Tuple{Array{Int64,1},Colon},0}:
 1  0  0  0  0
 0  2  0  0  0

julia> X[[1,2], [1,2]]
2-dimensional AxisArray{Int64,2,...} with axes:
    :c1, [:one,:two]
    :c2, [-5,-4]
And data, a 2x2 SubArray{Int64,2,NDSparseData.NDSparse{Int64,2,Tuple{Int64,Int64},Tuple{Array{Int64,1},Array{Int64,1}},Array{Int64,1}},Tuple{Array{Int64,1},Array{Int64,1}},0}:
 1  0
 0  2
````

And some other generic methods work on `NDSparse`:

````julia
julia> A = NDSparse(sprand(5,5,0.5))
NDSparse{Float64,2,Tuple{Int64,Int64}}:
 (1,1) => 0.4464544946933542
 (1,2) => 0.22752199955534458
 (1,4) => 0.5381053980698305
 (2,1) => 0.059329551148960435
 (2,3) => 0.06320603590648721
 (2,4) => 0.6675226471816889
 (2,5) => 0.863660157404958
 (3,3) => 0.07382710164164186
 (3,4) => 0.7526547391893967
 (3,5) => 0.7023654325273427

julia> collect(A)
5×5 Array{Float64,2}:
 0.446454   0.227522   0.0        0.538105  0.0     
 0.0593296  0.0        0.063206   0.667523  0.86366 
 0.0        0.0        0.0738271  0.752655  0.702365
 0.0        0.0800303  0.330118   0.465113  0.712696
 0.0        0.253965   0.0396619  0.367096  0.0     

julia> collect(A')
5×5 Array{Float64,2}:
 0.446454  0.0593296  0.0        0.0        0.0      
 0.227522  0.0        0.0        0.0800303  0.253965 
 0.0       0.063206   0.0738271  0.330118   0.0396619
 0.538105  0.667523   0.752655   0.465113   0.367096 
 0.0       0.86366    0.702365   0.712696   0.0      

julia> collect(A * A)
5×5 Array{Float64,2}:
 0.21282     0.144643   0.192019  0.642396  0.580007
 0.0264879   0.28626    0.259282  0.707018  0.520134
 0.0         0.238611   0.281772  0.663472  0.588267
 0.00474816  0.218223   0.211239  0.779845  0.632466
 0.0150676   0.0293788  0.140165  0.37012   0.508824
````

tanmaykm:
Does this look useful and good to merge?

Also, what was the thinking behind the `Dimension` and `Interval` types in `utils.jl`? They are currently unused, but look like something meant for indexing.

ViralBShah:
I think so.

tanmaykm:
Ok. Will merge tomorrow if there are no concerns.

JeffBezanson:
This completely changes the meaning of an NDSparseData. Previously, it supported any ordered types as indexes, e.g. strings or floats. Now it needs to have a fixed integer index space. I don't see how this is supposed to work. Maybe we can compute `size` as the number of unique elements in each dimension? Then it would at least be automatic.

tanmaykm:
Yes, this makes `NDSparse` more like `Array` than a `Dict`. I felt it was useful in some cases, while retaining indexing with cell values.

What if we set the default index space to `typemax(Int)` in each dimension, then there will be no need to recompute the size on each change.

tanmaykm:
But show on repl fails when a dim is `typemax(Int)`:

````
julia> A = NDSparse((typemax(Int),typemax(Int)), [1,2,3,4,5], [1,2,3,4,5], [1,2,3,4,5])
9223372036854775807×9223372036854775807 NDSparseData.NDSparse{Int64,2,Tuple{Int64,Int64},Tuple{Array{Int64,1},Array{Int64,1}},Array{Int64,1}}:
Error showing value of type NDSparseData.NDSparse{Int64,2,Tuple{Int64,Int64},Tuple{Array{Int64,1},Array{Int64,1}},Array{Int64,1}}:
ERROR: invalid Array size
 in vcat(::UnitRange{Int64}) at ./range.jl:750
 in print_matrix(::IOContext{Base.Terminals.TTYTerminal}, ::NDSparseData.NDSparse{Int64,2,Tuple{Int64,Int64},Tuple{Array{Int64,1},Array{Int64,1}},Array{Int64,1}}, ::String, ::String, ::String, ::String, ::String, ::String, ::Int64, ::Int64) at ./show.jl:1342
 in print_matrix(::IOContext{Base.Terminals.TTYTerminal}, ::NDSparseData.NDSparse{Int64,2,Tuple{Int64,Int64},Tuple{Array{Int64,1},Array{Int64,1}},Array{Int64,1}}, ::String, ::String, ::String) at ./show.jl:1330
 in #showarray#167(::Bool, ::Function, ::IOContext{Base.Terminals.TTYTerminal}, ::NDSparseData.NDSparse{Int64,2,Tuple{Int64,Int64},Tuple{Array{Int64,1},Array{Int64,1}},Array{Int64,1}}, ::Bool) at ./show.jl:1565
 in display(::Base.REPL.REPLDisplay{Base.REPL.LineEditREPL}, ::MIME{Symbol("text/plain")}, ::NDSparseData.NDSparse{Int64,2,Tuple{Int64,Int64},Tuple{Array{Int64,1},Array{Int64,1}},Array{Int64,1}}) at ./REPL.jl:114
...
````

because `collect(indices(A, 1))` fails:
````
ERROR: invalid Array size
 in vcat(::UnitRange{Int64}) at ./range.jl:750
 in collect(::UnitRange{Int64}) at ./range.jl:760
 in eval(::Module, ::Any) at ./boot.jl:234
 in macro expansion at ./REPL.jl:92 [inlined]
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:46
````

which shouldn't, right?

JeffBezanson:
I just don't think NDSparse can support `size`. While there are some possible interpretations, they wouldn't be useful to most consumers of AbstractArrays. If an array function sees a size of `typemax(Int)`, it might try to iterate over that range, or it might just bail out, neither of which are useful. Similarly, `indices` assumes that there is data at each point in the product of the ranges it returns.

I think the most pressing problem is the constructor. In my use cases, it's easy and natural to construct an NDSparse from some columns, but I have no idea what to pass for the size tuple. If we want to keep `size`, I think the constructor should look at the indexes and determine the size automatically if the indexes are integers, and otherwise leave the size unset. Then when `size` is called, if the size is unset we can throw an error. Of course I find that depressingly complex and would rather just leave it out though.

tanmaykm:
I am not doing much with `AxisArrays` and `NDSparse` as originally intended, and I don't think any one else is dependent on this as of now.

It's probably best to revert this.

PR #10, opened 2016-05-25
RFC: Fixes + Benchmarking

pranavtbhat:
Fix for https://github.com/JuliaComputing/NDSparseData.jl/issues/8 and speedup for `select`.

For the benchmarking framework refer doc.md. I've written benchmarks for getindex, setindex!, flush!, merge and the queries. Should I cover anything else?


src/NDSparseData.jl
```diff
@@ -209,6 +209,7 @@ function _getindex(t::NDSparse, idxs)
     end
     I = t.indexes
     lI = length(I)
+    map(sort!, idxs)
```
JeffBezanson: I'm worried about the performance of this. Might be better to check `issorted` and throw an error instead? This also won't support immutable indexes, such as ranges or `:`.

JeffBezanson:
Cool! Another good thing to benchmark is construction, which includes a sorting step that can be slow. Should test a couple kinds of data, such as all numbers vs. mix of numbers and strings, etc.

src/NDSparseData.jl
pranavtbhat: Will this work better?
```julia
for idx in idxs
   isa(idx, AbstractVector) && (issorted(idx) || error("indexes must be sorted for ranged/vector indexing"))
end
```

pranavtbhat:
Added construction to benchmarking. Also replaced `sort!` with `issorted` for indexes of type `AbstractVectors` in vector indexing.

PR #24, opened 2016-08-05
add NamedTuple support for named columns and struct-of-arrays data

JeffBezanson:
The approach is (1) use `Columns` for both indexes and data, (2) bring in NamedTuples. Example:

```
julia> NDSparse(rand(3), Columns(open = rand(3), close = rand(3)))
NDSparse{NDSparseData._NT_openclose{Float64,Float64},Tuple{Float64}}:
 (0.07459099981805961,) => (open => 0.39695970629326216, close => 0.8768181865079363)
 (0.23068405270544878,) => (open => 0.21901570490914923, close => 0.9516909318576521)
 (0.8329699883751216,) => (open => 0.23058501858024938, close => 0.8031837612378654)
```

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,2 @@
-julia 0.4
+julia 0.5-
+NamedTuples
--- a/src/NDSparseData.jl
+++ b/src/NDSparseData.jl
@@ -1,5 +1,7 @@
 module NDSparseData
 
+using NamedTuples
+
 import Base:
     show, eltype, length, getindex, setindex!, ndims, map, convert,
     ==, broadcast, broadcast!, empty!, copy, similar, sum, merge,
@@ -10,7 +12,7 @@ export NDSparse, flush!, aggregate!, where, pairs, convertdim
 include("utils.jl")
 include("columns.jl")
 
-immutable NDSparse{T, D<:Tuple, C<:Tuple, V<:AbstractVector}
+immutable NDSparse{T, D<:Union{Tuple,NamedTuple}, C<:Tuple, V<:AbstractVector}
     index::Columns{D,C}
     data::V
 
--- a/src/NDSparseData.jl
+++ b/src/NDSparseData.jl
@@ -73,21 +73,47 @@ showarray(io::IO, t::NDSparse) = show(io, t)
 
 function show{T,D<:Tuple}(io::IO, t::NDSparse{T,D})
     flush!(t)
-    print(io, "NDSparse{$T,$D}:")
-    if isa(t.index.columns, NamedTuple)
-        print(io, "\n ")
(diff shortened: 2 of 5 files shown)


ViralBShah:
Extracting a vector of a particular data field will still be type unstable, right? The challenge here is also not being able to use our vectorization operators effectively. @tanmaykm experimented with this, I believe - so would be good to hear what he may have to say.

JeffBezanson:
Not necessarily --- we could use a named tuple for the tuple of columns, and then you can get a whole column with `data.col` which is type-inferrable.

A larger problem here is that the mental model of arrays requires that the array be *of* a particular thing. If you have an array of things of type X, operations that make sense on X can be lifted to the array. It also allows names of data to be first-class variable names in the language. Together this enables vectorized statements like `price1 .- price2`. But if the `price1` object contains multiple data columns, then the actual data is no longer first-class --- you have no way of naming the specific data you want within the language --- and many problems ensue.

tanmaykm:
Viral is referring to was what we saw here: https://github.com/tanmaykm/notebooks/blob/master/stocks/compare_df_like.ipynb
But it was not using named tuples then.

JeffBezanson:
Yes, there is always potential for type-unstable code whenever you index into a structure containing more than one type. True of simple tuples as well.

Actually, this PR is mostly independent of those issues, because all it really does is generalize the `Columns` structure so it can produce NamedTuples as well as Tuples.

What I'm working on now is allowing the tuple of columns to itself be a NamedTuple, which naturally provides for named dimensions. Almost all the code can stay the same, since the only difference is that columns can be indexed by symbol as well as integer index. So far it feels like the right approach.

JeffBezanson:
Added named columns. Demo:

```
julia> hi = NDSparse(Columns(city=[fill("New York",3); fill("Boston",3)],
                             date=repmat(Date(2016,7,6):Date(2016,7,8), 2)),
                             [91,89,91,95,83,76])
NDSparse{Int64,Tuple{String,Date}}:
 (:city,:date)
 ("Boston",2016-07-06) => 95
 ("Boston",2016-07-07) => 83
 ("Boston",2016-07-08) => 76
 ("New York",2016-07-06) => 91
 ("New York",2016-07-07) => 89
 ("New York",2016-07-08) => 91

julia> select(hi, :city => x->x=="Boston")
NDSparse{Int64,Tuple{Array{String,1},Array{Date,1}}}:
 (:city,:date)
 ("Boston",2016-07-06) => 95
 ("Boston",2016-07-07) => 83
 ("Boston",2016-07-08) => 76
```

Currently this requires my fork of NamedTuples.

JeffBezanson:
Updates to NamedTuples are now on master there, so I'll merge this.

PR #30, opened 2017-02-14
wip 0.6 fixes

### LINKED ISSUE #29: Columns with name argument fails first time on 0.6-
```jl
julia> using IndexedTables

julia> Columns(x = [1])
ERROR: MethodError: Cannot `convert` an object of type Array{Int64,1} to an object of type NamedTuples
._NT_x{Array{Int64,1}}
This may have arisen from a call to the constructor NamedTuples._NT_x{Array{Int64,1}}(...),
since type constructors fall back to convert methods.
Stacktrace:
 [1] NamedTuples._NT_x{Array{Int64,1}}(::Array{Int64,1}) at ./sysimg.jl:24
 [2] #Columns#14(::Array{Symbol,1}, ::Type{T}, ::Array{Int64,1}, ::Vararg{Array{Int64,1},N}) at /home/
andreas/.julia/v0.6/IndexedTables/src/columns.jl:28
 [3] (::Core.#kw#Type)(::Array{Any,1}, ::Type{IndexedTables.Columns}, ::Array{Int64,1}) at ./<missing>
:0
 [4] #Columns#15(::Array{Any,1}, ::Type{T}) at /home/andreas/.julia/v0.6/IndexedTables/src/columns.jl:
32
 [5] (::Core.#kw#Type)(::Array{Any,1}, ::Type{IndexedTables.Columns}) at ./<missing>:0

julia> Columns(x = [1])
1-element IndexedTables.Columns{NamedTuples._NT_x{Int64},NamedTuples._NT_x{Array{Int64,1}}}:
 (x => 1)
```
The build is
```jl
julia> versioninfo()
Julia Version 0.6.0-dev.2010
Commit 7c34d69* (2017-01-09 15:23 UTC)
Platform Info:
  OS: Linux (x86_64-redhat-linux)
  CPU: Intel(R) Xeon(R) CPU E5-2697 v2 @ 2.70GHz
  WORD_SIZE: 64
  BLAS: libmkl_rt
  LAPACK: libmkl_rt
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, ivybridge)
```

shashi:
Trying to fix #29 

there were a few easy issues in NamedTuples to get done first: https://github.com/blackrock/NamedTuples.jl/compare/master...shashi:0.6-compat?expand=1

then this happened:
```
julia> NDSparse(Columns(x = rand(4), y = rand(4)), Columns(observation = rand(1:2, 4), confidence=rand(4)));
ERROR: MethodError: no method matching NamedTuples._NT_xy{Array{Float64,1},Array{Float64,1}}(::Array{Float64,1}, ::Array{Float64,1})
The applicable method may be too new: running in world age 21611, while current world is 21613.
Stacktrace:
 [1] #Columns#14(::Array{Symbol,1}, ::Type{T} where T, ::Array{Float64,1}, ::Vararg{Array{Float64,1},N} where N) at /home/shashi/.julia/v0.6/IndexedTables/src/columns.jl:28
 [2] (::Core.#kw#Type)(::Array{Any,1}, ::Type{IndexedTables.Columns}, ::Array{Float64,1}, ::Array{Float64,1}) at ./<missing>:0
 [3] #Columns#15(::Array{Any,1}, ::Type{T} where T) at /home/shashi/.julia/v0.6/IndexedTables/src/columns.jl:32
 [4] (::Core.#kw#Type)(::Array{Any,1}, ::Type{IndexedTables.Columns}) at ./<missing>:0
```

I figured I can evade the world error by not using eval in the Columns constructor, I tried using a generated function on a tuple of Vals denoting the fields as in this diff. But! There's an eval in the NamedTuples constructor... So I get `cannot eval in a generated function` here.

I tried to employ the same plan of using `Val{Sym}`s and using `@generated` in named tuples creation, but then realized I can't create a type inside function scope so that wouldn't work.

I don't know if there's anything else that can be done to get this back up and running. These are some options I see:

Option 1: we can switch to `FixedSizeDict`s or something similar, but that comes with caveats:

- Indexing with symbols is type unstable but might get better with @vtjnash's Union type work.
- Indexing with `Val` is fine, but `x[Val{:y}]` is far from the ergonomics of `x.y`.
- There's a macro sugar `@get x.y`, but still relatively not nice.

Option 2: with a little help from Julia the language: (any one of)

- first-class named tuples support (favorite option)
- dot-overloading (least favorite)
- allow types in function scope

cc @JeffBezanson @vtjnash

### DIFF
--- a/src/IndexedTables.jl
+++ b/src/IndexedTables.jl
@@ -35,7 +35,7 @@ Keyword arguments:
 * `presorted::Bool`: If true, the indices are assumed to already be sorted and no sorting is done.
 * `copy::Bool`: If true, the storage for the new array will not be shared with the passed indices and data. If false (the default), the passed arrays will be copied only if necessary for sorting. The only way to guarantee sharing of data is to pass `presorted=true`.
 """
-function NDSparse{T,D,C}(I::Columns{D,C}, d::AbstractVector{T}; agg=nothing, presorted=false, copy=false)
+function NDSparse{T<:AbstractVector, D,C}(I::Columns{D,C}, d::T; agg=nothing, presorted=false, copy=false)
     length(I) == length(d) || error("index and data must have the same number of elements")
     # ensure index is a `Columns` that generates tuples
     dt = D
@@ -56,7 +56,7 @@ function NDSparse{T,D,C}(I::Columns{D,C}, d::AbstractVector{T}; agg=nothing, pre
             d = Base.copy(d)
         end
     end
-    nd = NDSparse{T,dt,C,typeof(d)}(I, d, similar(I,0), similar(d,0))
+    nd = NDSparse{eltype(T),dt,C,typeof(d)}(I, d, similar(I,0), similar(d,0))
     agg===nothing || aggregate!(agg, nd)
     return nd
 end
@@ -68,7 +68,10 @@ Construct an NDSparse array from columns. The last argument is the data column, 
 """
 function NDSparse(columns...; names=nothing, rest...)
     keys, data = columns[1:end-1], columns[end]
-    NDSparse(Columns(keys..., names=names), data; rest...)
+    keycols = Columns(keys..., names=names)
+    @show typeof(keycols) typeof(data)
+    @show rest
+    NDSparse(keycols, data; rest...)
 end
 
 similar(t::NDSparse) = NDSparse(similar(t.index), empty!(similar(t.data)))
--- a/src/columns.jl
+++ b/src/columns.jl
@@ -19,13 +19,22 @@ immutable Columns{D<:Tup, C<:Tup} <: AbstractVector{D}
     end
 end
 
+@generated function _named_columns(cols::Tuple, names::Tuple)
+    name_syms = [n.parameters[1] for n in names.parameters]
+    eltypes = map(eltype, cols.parameters)
+    coltypes = cols.parameters
+    quote
+        dt = @NT($(name_syms...)){$(eltypes...)}
+        ct = @NT($(name_syms...)){$(coltypes...)}
+        Columns{dt,ct}(ct($cols...))
+    end
+end
+
 function Columns(cols::AbstractVector...; names::Union{Vector{Symbol},Tuple{Vararg{Symbol}},Void}=nothing)
     if isa(names, Void)
         Columns{eltypes(typeof(cols)),typeof(cols)}(cols)
     else
-        dt = eval(:(@NT($(names...)))){map(eltype, cols)...}
-        ct = eval(:(@NT($(names...)))){map(typeof, cols)...}
-        Columns{dt,ct}(ct(cols...))
+        _named_columns(cols, map(x->Val{x}(), (names...)))
     end
 end
 


vtjnash:
> and using `@generated` in named tuples creation

right. the capabilities of `@generated` are a strict subset of all other functions, while the capabilities of `eval` is a strict superset of all other functions.

because this uses `eval`, it's already entirely opaque to the compiler. AFAICT, because the design of NamedTuples correctly ensures that this will not happen, there is no way to express an operation like this – all operations must be written as explicit macros in the source code.

shashi:
Thanks for clarifying. So then we need to pick one of the options from that list or something else.

We could have `@NDSparse(...)` as well as another option.

ViralBShah:
It's best to have a regular weekly JuliaDB call to discuss all this. Tough to do such design discussions on github.

shashi:
I'm not sure what I'm doing wrong 
```julia
NDSparse([1], [2])
```
goes into an infinite loop

It enssentially calls `NDSparse(Columns([1]), [2])` which instead of calling [this method](https://github.com/JuliaComputing/IndexedTables.jl/pull/30/commits/e6a73ba9a8f2336a67f27405af7bd6a86579a036#diff-e450c03168de7e91daf296bc92c152cfR38) calls `NDSparse(cols...)`

I added debug statements to print the types of the arguments inside the `NDSparse(cols...)` to verify that the types are indeed ok to be dispatched on the other method.

Works fine on 0.5.

```julia
julia> using IndexedTables

julia> NDSparse(Columns([1]), [2])
typeof(keycols) = IndexedTables.Columns{Tuple{Tuple{Int64}},Tuple{IndexedTables.Columns{Tuple{Int64},Tuple{Array{Int64,1}}}}}
typeof(data) = Array{Int64,1}
rest = Any[]
typeof(keycols) = IndexedTables.Columns{Tuple{Tuple{Tuple{Int64}}},Tuple{IndexedTables.Columns{Tuple{Tuple{Int64}},Tuple{IndexedTables.Columns{Tuple{Int64},Tuple{Array{Int64,1}}}}}}}
typeof(data) = Array{Int64,1}
rest = Any[]
typeof(keycols) = IndexedTables.Columns{Tuple{Tuple{Tuple{Tuple{Int64}}}},Tuple{IndexedTables.Columns{Tuple{Tuple{Tuple{Int64}}},Tuple{IndexedTables.Columns{Tuple{Tuple{Int64}},Tuple{IndexedTables.Columns{Tuple{Int64},Tuple{Array{Int64,1}}}}}}}}}
typeof(data) = Array{Int64,1}
rest = Any[]
typeof(keycols) = IndexedTables.Columns{Tuple{Tuple{Tuple{Tuple{Tuple{Int64}}}}},Tuple{IndexedTables.Columns{Tuple{Tuple{Tuple{Tuple{Int64}}}},Tuple{IndexedTables.Columns{Tuple{Tuple{Tuple{Int64}}},Tuple{IndexedTables.Columns{Tuple{Tuple{Int64}},Tuple{IndexedTables.Columns{Tuple{Int64},Tuple{Array{Int64,1}}}}}}}}}}}
typeof(data) = Array{Int64,1}
rest = Any[]
typeof(keycols) = IndexedTables.Columns{Tuple{Tuple{Tuple{Tuple{Tuple{Tuple{Int64}}}}}},Tuple{IndexedTables.Columns{Tuple{Tuple{Tuple{Tuple{Tuple{Int64}}}}},Tuple{IndexedTables.Columns{Tuple{Tuple{Tuple{Tuple{Int64}}}},Tuple{IndexedTables.Columns{Tuple{Tuple{Tuple{Int64}}},Tuple{IndexedTables.Columns{Tuple{Tuple{Int64}},Tuple{IndexedTables.Columns{Tuple{Int64},Tuple{Array{Int64,1}}}}}}}}}}}}}
typeof(data) = Array{Int64,1}
rest = Any[]
typeof(keycols) = IndexedTables.Columns{Tuple{Tuple{Tuple{Tuple{Tuple{Tuple{Tuple{Int64}}}}}}},Tuple{IndexedTables.Columns{Tuple{Tuple{Tuple{Tuple{Tuple{Tuple{Int64}}}}}},Tuple{IndexedTables.Columns{Tuple{Tuple{Tuple{Tuple{Tuple{Int64}}}}},Tuple{IndexedTables.Columns{Tuple{Tuple{Tuple{Tuple{Int64}}}},Tuple{IndexedTables.Columns{Tuple{Tuple{Tuple{Int64}}},Tuple{IndexedTables.Columns{Tuple{Tuple{Int64}},Tuple{IndexedTables.Columns{Tuple{Int64},Tuple{Array{Int64,1}}}}}}}}}}}}}}}
typeof(data) = Array{Int64,1}
```

vtjnash:
There's a bug open on Julia by Jarrett that looks very similar to that.

shashi:
If you are referring to https://github.com/JuliaLang/julia/issues/20200 this seems to be similar but worse because IndexedTables doesn't call __precompile__()

JeffBezanson:
Working on it; looks like a specificity issue.

vtjnash:
> We could have @NDSparse(...) as well as another option.

Right. That's what NamedTuples does and what it expects callers to do.

JeffBezanson:
I have an interesting fix for the NamedTuples issue. Since NamedTuples are like tuples (constructor arguments always just directly become field values) you can reuse a single generic constructor for all NamedTuple types, and there is no need to add new methods to `Type` for each NamedTuple type. Since no methods are added, it becomes possible to use NamedTuple types constructed at runtime as before. In practice, it helps to pre-generate specializations for small argument counts, or the whole thing could be a generated function, but that's not strictly necessary.

shashi:
> there is no need to add new methods to Type for each NamedTuple type. Since no methods are added, it becomes possible to use NamedTuple types constructed at runtime as before.

Sounds interesting, but I'm not sure what this means... How do you generate new types without adding a method to `Type`? Will we still be able to use `nt.foo` syntax? Or will it be `nt[:foo]`?

JeffBezanson:
See here: https://github.com/blackrock/NamedTuples.jl/pull/18
It's fully backwards compatible and I think should work for all uses of NamedTuples without changing any code. The key is a dirty trick to prevent generation of default constructors.

PR #32, opened 2017-03-01
RFC: Add a kw-arg method to aggregate_vec

tshort:
This makes it a bit nicer to perform multiple aggregations on groups.

```julia
tbl = NDSparse(rand(1:3, 10), rand(1:2, 10), rand(10))
aggregate_vec(tbl, max = maximum, q75 = x->quantile(x, 0.75))
```

I included two versions, one with Pairs and one with keyword arguments. I like the keyword arg version better.

### DIFF
--- a/src/query.jl
+++ b/src/query.jl
@@ -185,6 +185,15 @@ function aggregate_vec(f, x::NDSparse)
     NDSparse(idxs, data, presorted=true, copy=false)
 end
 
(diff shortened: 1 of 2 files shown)


JeffBezanson:
Yes, let's go with the keyword argument version. We should also refactor to have both methods call a helper function that accepts `names` and `functions` arguments, to avoid duplicating code.

tshort:
I think I made the changes you requested, @JeffBezanson.

PR #38, opened 2017-03-31
Remove map for NamedTuple

davidanthoff:
This should be the only thing left to do so that the package can be registered (JuliaLang/METADATA.jl#8476).

The REQUIRE bound on ``NamedTuples`` is not ideal, the proper requirement would be ``2.1.0`` if you are on ``2.x``, or ``3.0.2`` if you are on ``3.x``. But I don't think that kind of requirement can be written, so this is probably the best we can do.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.5
 Compat 0.19
-NamedTuples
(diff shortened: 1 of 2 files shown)


davidanthoff:
Thanks!

PR #40, opened 2017-04-03
add `agg` option to merge to combine equal indices

shashi:


shashi:
Updated to add an `agg=nothing` case which leaves repeating values in the output. useful for calling `aggregate_vec` on merged output.

Should there be a `vecagg` kwarg as well? It seems like redundant API given its implementation would just call `aggregate_vec!`.

src/join.jl
```diff
@@ -170,32 +170,48 @@ function count_overlap{D}(I::Columns{D}, J::Columns{D})
...
-                if c==0; i += 1; end
                 j += 1
+            elseif c < 0
+                K[k] = I[i]
+                data[k] = x.data[i]
+                i += 1
             else
                 K[k] = I[i]
                 data[k] = x.data[i]
+                if isa(agg, Void)
```
shashi: would this branch be removed at compile time?

JeffBezanson: Yes, as would `agg === nothing`.

PR #41, opened 2017-04-06
RFC/WIP: mapslices

shashi:
a first attempt at mapslices:

scalar version (i.e. if map function doesn't return an `IndexedTable`)

```julia
julia> using IndexedTables

julia> x=IndexedTable(Columns(a=[1,1,1,2,2],b=[1,2,3,1,2]),[1,2,3,4,5])
a  b │ 
─────┼──
1  1 │ 1
1  2 │ 2
1  3 │ 3
2  1 │ 4
2  2 │ 5

julia> mapslices(println, x, [1])
a  b │ 
─────┼──
1  1 │ 1
2  1 │ 4
a  b │ 
─────┼──
1  2 │ 2
2  2 │ 5
a  b │ 
─────┼──
1  3 │ 3
b │ 
──┼────────
1 │ nothing
2 │ nothing
3 │ nothing

julia> mapslices(println, x, [:b])
a  b │ 
─────┼──
1  1 │ 1
1  2 │ 2
1  3 │ 3
a  b │ 
─────┼──
2  1 │ 4
2  2 │ 5
a │ 
──┼────────
1 │ nothing
2 │ nothing
```
non-scalar version (i.e. function returns an IndexedTable):
```julia
julia> r=Ref(0);

julia> mapslices(x, [:a]) do slice
           println("input:")
           println(slice)
           r[] += 1
           n = length(slice)
           output = IndexedTable(Columns(c=[1:n;]), [r[] for i=1:n])
           println("output:")
           println(output)
           output
       end
input:
a  b │ 
─────┼──
1  1 │ 1
2  1 │ 4
output:
c │ 
──┼──
1 │ 1
2 │ 1
input:
a  b │ 
─────┼──
1  1 │ 1
2  1 │ 4
output:
c │ 
──┼──
1 │ 2
2 │ 2
input:
a  b │ 
─────┼──
1  2 │ 2
2  2 │ 5
output:
c │ 
──┼──
1 │ 3
2 │ 3
input:
a  b │ 
─────┼──
1  3 │ 3
output:
c │ 
──┼──
1 │ 4
b  c │ 
─────┼──
1  1 │ 1
1  1 │ 2
1  2 │ 1
1  2 │ 2
2  1 │ 3
2  2 │ 3
3  1 │ 4

julia> r=Ref(0);

julia> mapslices(x, [:b]) do slice
           println("input:")
           println(slice)
           r[] += 1
           n = length(slice)
           output = IndexedTable(Columns(c=[1:n;]), [r[] for i=1:n])
           println("output:")
           println(output)
           output
       end
input:
a  b │ 
─────┼──
1  1 │ 1
1  2 │ 2
1  3 │ 3
output:
c │ 
──┼──
1 │ 1
2 │ 1
3 │ 1
input:
a  b │ 
─────┼──
1  1 │ 1
1  2 │ 2
1  3 │ 3
output:
c │ 
──┼──
1 │ 2
2 │ 2
3 │ 2
input:
a  b │ 
─────┼──
2  1 │ 4
2  2 │ 5
output:
c │ 
──┼──
1 │ 3
2 │ 3
a  c │ 
─────┼──
1  1 │ 1
1  1 │ 2
1  2 │ 1
1  2 │ 2
1  3 │ 1
1  3 │ 2
2  1 │ 3
2  2 │ 3
```

basically dimensions of the returned slices are tacked on to the end of the remaining indices of the table.

Some questions:

- what should we do if some dimensions returned from the mapping function are common to the remaining dimensions? Right now they get a new name: `a` becomes `a_1` and the rest is handled similarly
- should we drop the sliced dimension in the input? It seems useful, but may be redundant.
- should this function take an aggregate function as well?

```
julia> mapslices(x->rand(), rand(2,2), 1)
1×2 Array{Float64,2}:
 0.302282  0.403092

julia> mapslices(x->rand(2), rand(2,2), 1)
2×2 Array{Float64,2}:
 0.759656  0.0750492
 0.801441  0.273785 

julia> mapslices(x->rand(2,2), rand(2,2), 1)
ERROR: DimensionMismatch("tried to assign 2 elements to 1 destinations")
```

what's the equivalent of this??

cc @simonbyrne

### DIFF
--- a/src/query.jl
+++ b/src/query.jl
@@ -284,3 +284,113 @@ function reducedim_vec(f, x::IndexedTable, dims)
 end
 
 reducedim_vec(f, x::IndexedTable, dims::Symbol) = reducedim_vec(f, x, [dims])
+
+function dedup_names(ns)
+    count = Dict{Symbol,Int}()
+    for n in ns
+        if haskey(count, n)
+            count[n] += 1
+        else
+            count[n] = 1
+        end
+    end
+
+    repeated = filter((k,v) -> v > 1, count)
+    for k in keys(repeated)
+        repeated[k] = 0
+    end
+    [haskey(repeated, n) ? Symbol(n, "_", repeated[n]+=1) : n for n in ns]
+end
+
+function mapslices(f, x::IndexedTable, dims; name = nothing)
+    iterdims = setdiff([1:ndims(x);], map(d->fieldindex(x.index.columns,d), dims))
+    idx = Any[Colon() for v in x.index.columns]
+
+    iter = sort(Columns(astuple(x.index.columns)[[iterdims...]]))
+
+    for j in 1:length(iterdims)
+        d = iterdims[j]
+        idx[d] = iter[1][j]
+    end
+    if length(idx) == length(iterdims)
+        idx[end] = vcat(idx[end]) # make last element a vector
+                                  # so that we don't do scalar indexing
+    end
+    y = f(x[idx...]) # Apply on first slice
+
+    if isa(y, IndexedTable)
+        # this means we need to concatenate outputs into a big IndexedTable
+        ns = vcat(dimlabels(x)[iterdims], dimlabels(y))
+        if !all(x->isa(x, Symbol), ns)
+            ns = nothing
+        else
+            ns = dedup_names(ns)
+        end
+        n = length(y)
+        index_first = similar(iter, n)
+        for j=1:n
+            @inbounds index_first[j] = iter[1]
+        end
+        index = Columns(index_first.columns..., astuple(y.index.columns)...; names=ns)
+        data = copy(y.data)
+        output = IndexedTable(index, data)
+        _mapslices_itable!(f, output, x, iter, iterdims, 2)
+    else
+        ns = dimlabels(x)[iterdims]
+        if !all(x->isa(x, Symbol), ns)
+            ns = nothing
+        end
+        index = Columns(iter[1:1].columns...; names=ns)
+        if name === nothing
+            output = IndexedTable(index, [y])
+        else
+            output = IndexedTable(index, Columns([y], names=[name]))
+        end
+        _mapslices_scalar!(f, output, x, iter, iterdims, 2, name!==nothing ? x->(x,) : identity)
+    end
+end
+
+function _mapslices_scalar!(f, output, x, iter, iterdims, start, coerce)
+    idx = Any[Colon() for v in x.index.columns]
+
+    for i = start:length(iter)
+        if i != 1 && roweq(iter, i-1, i) # We've already visited this slice
+            continue
+        end
+        for j in 1:length(iterdims)
+            d = iterdims[j]
+            idx[d] = iter[i][j]
+        end
+        y = f(x[idx...])
+        push!(output.index, iter[i])
+        push!(output.data, coerce(y))
+    end
+    output
+end
+
+function _mapslices_itable!(f, output, x, iter, iterdims, start)
+    idx = Any[Colon() for v in x.index.columns]
+    I = output.index
+    D = output.data
+    initdims = length(iterdims)
+
+    I1 = Columns(I.columns[1:initdims]) # filled from existing table
+    I2 = Columns(I.columns[initdims+1:end]) # filled from output tables
+
+    for i = start:length(iter)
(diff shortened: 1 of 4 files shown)


simonbyrne:
Looks good. However I also need it to work when mapping over all variables (i.e. when the slice is a singleton):
```julia
x=IndexedTable(Columns(a=[1,1,1,2,2],b=[1,2,3,1,2]),[1,2,3,4,5])
# works
mapslices(x, [:a]) do slice
    IndexedTable(Columns(c=[1,2]), [2,7])
end
# doesn't work 
mapslices(x, [:a,:b]) do slice
    IndexedTable(Columns(c=[1,2]), [2,7])
end
```

Also, as I mentioned on slack, a special case of these are `stack` and `unstack` (or `melt` and `cast` in dplyr speak).

shashi:
```julia
mapslices(x, [:a,:b]) do slice
    IndexedTable(Columns(c=[1,2]), [2,7])
end
```
It seems what you wanted was `mapslices(f, x, [])` ? (I've made it work) `mapslices(f, x, [:a, :b])` would pass the entire table to `f`... Is this confusing?

JeffBezanson:
Isn't `mapslices(f, x, [])` the same as `map`?

src/query.jl
```diff
@@ -282,3 +282,129 @@ function reducedim_vec(f, x::IndexedTable, dims)
 end
 
 reducedim_vec(f, x::IndexedTable, dims::Symbol) = reducedim_vec(f, x, [dims])
+
+function dedup_names(ns)
+    count = Dict{Symbol,Int}()
+    ns1 = copy(ns)
+    for i in length(ns):-1:1
+        if ns[i] in ns[1:i-1]
```
JeffBezanson: This is mostly redundant with the dict lookup. Could avoid the O(n^2) here.

src/query.jl
```diff
@@ -282,3 +282,129 @@ function reducedim_vec(f, x::IndexedTable, dims)
...
+        for k = (srcdims+1):srcdims+length(y.index.columns)
+            ldestcol = I.columns[k]
+            srccol = y.index.columns[k-srcdims]
+            @inbounds for j=l+1:l+n
+                ldestcol[j] = srccol[j-l]
+            end
+        end
+
+        @inbounds for j=l+1:l+n
+            D[j] = y.data[j-l]
```
JeffBezanson: Could these loops be replaced with `append!` ?

simonbyrne:
Ah, yes you're right, I got confused.

@JeffBezanson The problem is that `map(s -> IndexedTable(Columns(c=[1,1],d=[1,2]),[1,2]), x)` gives you an `IndexedTable` of `IndexedTable`s.

JeffBezanson:
Ah yes, that's tricky. `mapslices` with `[]` dimensions on normal `Array`s is actually equivalent to `map`, but dropping dimensions when indexing IndexedTables is not always natural, since the indices still have meaning. Scalar indexing of IndexedTables is currently kind of a special case. Would be nice to have a clean approach to this.

simonbyrne:
For some reason I'm getting an extra copy of the first row:
```
julia> using IndexedTables

julia> x=IndexedTable(Columns(a=[1,2,3,4,5]),Columns(b=[1,2,3,4,5]))
a │ b
──┼──
1 │ 1
2 │ 2
3 │ 3
4 │ 4
5 │ 5

julia> mapslices(x,()) do slice
                IndexedTable(Columns(z=[7]),
                             Columns(y=[1]))
       end
a  z │ y
─────┼──
1  7 │ 1
1  7 │ 1
2  7 │ 1
3  7 │ 1
4  7 │ 1
5  7 │ 1
```

src/query.jl
shashi: Fixed

src/query.jl
shashi: Replaced... I was trying to avoid creating new `Columns` for every slice just to append... worked around that by splitting `I` into 2 sets `Columns` in the beginning. I also added an `append_n!` to utils.jl, this is 3-4x faster than using `append!(x, Iterators.repeated(v,n))`

shashi:
> dropping dimensions when indexing IndexedTables is not always natural

True, I think then it makes sense then that in `mapslices` the slices also have the dimensions that are dropped in mapslices on arrays.

PR #43, opened 2017-05-11
promote_similar for output data columns in merge

shashi:
Sometimes some chunks may not contain missing values in some columns, but may in others. Or some chunks may be all Integers and some may have floats. In such cases we need to promote the output columns type while doing a `merge`.


shashi:
Updated to also promote index columns similarly. Travis failure seems to be due to missing native dependency on OSX.

PR #48, opened 2017-05-31
use IntervalSets for indexing

shashi:
There are two optimizations here:

1) define `_in` for `Range` which was getting caught by `_in(::AbstractVector)` method and doing a slow lookup
2) Try to omit unnecessary columns (for which `_in` will always be true) in the second-pass check to see if indices fall inside the specified query.

Also, we now reexport `..` from IntervalSets which makes for a really nice indexing experience for Dates (you don't have to mention timestep of 1 millisecond, you can even give a Date range for a DateTime column and it will be closed range which includes the whole 2 days!). In general indexing with ClosedInterval is also faster than with ranges.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,3 +2,4 @@ julia 0.5 0.7-
 Compat 0.19
 NamedTuples 2.1.0
 PooledArrays
+IntervalSets
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -1,5 +1,9 @@
 # getindex
 
+using IntervalSets
+
+export ClosedInterval, ..
+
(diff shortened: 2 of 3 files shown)


shashi:
indexing benchmark results versus `s/benchmarks`
```
const-const		+0.52%	invariant
colon-const		-1.35%	invariant
colon-colon		-77.82%	improvement
interval-colon		-88.23%	improvement
interval-const		-85.42%	improvement
const-interval		+4.11%	invariant
range-colon		-42.22%	improvement
range-const		-43.57%	improvement
interval-interval	-71.29%	improvement
range-interval		-36.64%	improvement
vector-interval		+1.23%	invariant
vector-vector		-0.07%	invariant
```

Issue #52, opened 2017-06-29
map and naming the resulting column


tcovert:
Is there a way to assign a name to the resulting column that map creates?  I would have thought `@NT()` would do this but it seems not to exactly work:

```
julia> tbl = IndexedTable(Columns(idx = collect(1:10)), Columns(vals = randn(10)))
julia> map(x->abs.(x.vals), tbl)
julia> map(x->@NT(aval = abs.(x.vals)), tbl)
```
the first map statement gives the output I'd expect but no name for the data column.  the second map statement does have a name, but it gives me an array of named tuples as opposed to an array of values.

thanks in advance for any suggestions.

shashi:
I agree that when the function returns a tuple or named tuple, `map` should construct new `Columns`...  I've made a PR for this #54

shashi:
Closed by #57

PR #54, opened 2017-07-10
map(f, t): create Columns when `f` returns tuple / named tuple

shashi:
Turned out to be a bit complicated...

addresses #52

### DIFF
--- a/src/IndexedTables.jl
+++ b/src/IndexedTables.jl
@@ -192,7 +192,84 @@ end
 
 # map and convert
 
-map(f, x::IndexedTable) = IndexedTable(copy(x.index), map(f, x.data), presorted=true)
+function _map_onerror!(i, val, f, out, inp)
+    n = length(inp)
+
+    T = promote_type(eltype(out), typeof(val))
+    if (T <: Tup) && (length(val) == nfields(T) && !any(t->t<:Vararg, T.parameters))
+        # There are just as many columns but the types of the
+        # elements has changed
+        cols = map(t->Array{t}(n), (T.parameters...))
+        promoted = Columns{T, typeof(cols)}(cols)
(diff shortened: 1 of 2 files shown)


davidanthoff:
Should a map that returns a ``Pair{NamedTuple{...}, NamedTuple{...}}`` return an ``IndexedTable`` where the columns in the key of the pair make up the index, and the columns in the value make up the data columns? This would essentially be #55, but it would use a normal ``map`` as the API.

In some sense, all we would need is a constructor for ``IndexedTable`` that takes an iterator of that element type. That is something I wanted to add for some while: davidanthoff/IterableTables.jl#26

shashi:
> Should a map that returns a Pair{NamedTuple{...}, NamedTuple{...}} return an IndexedTable where the columns in the key of the pair make up the index, and the columns in the value make up the data columns?

I don't think so... `map` should only work on values. You could have output keys that collide allowing you to implement `filter` with map which seems wrong.

>  a constructor for IndexedTable that takes an iterator of that element type

Good idea to add this... One could even convert any Associative to an IndexedTable just by calling this constructor.

shashi:
An alternative approach using `Base.promote_op` is in #57

shashi:
```julia
t = IndexedTable([1,2,3], Columns(x=[4,5,6]))
y = [1, 1//2, "x"]
t4 = map(x->(x.x, y[x.x-3]), t)
```
This doesn't work with the alternative approach as `promote_op` has difficulty figuring out the output type. Creating a function with return type annotation also didn't work...

shashi:
> has difficulty figuring out the output type

Turns out it was throwing away the fact that the results are tuples when any field is UnionAll... Fixed in #57

PR #57, opened 2017-07-11
Allow arbitrary return types in map and broadcast

shashi:
Supersedes #54

### DIFF
--- a/src/join.jl
+++ b/src/join.jl
@@ -424,10 +424,13 @@ If `dimmap` is not specified, it is determined automatically using index column
 names and types.
 """
 function broadcast(f::Function, A::IndexedTable, B::IndexedTable; dimmap=nothing)
+    out_T = Base.promote_op(f, eltype(A), eltype(B))
+    out = IndexedTable(similar(typeof(A.index), 0), similar(arrayof(out_T), 0))
+
     if ndims(B) > ndims(A)
-        _broadcast!((x,y)->f(y,x), similar(B), B, A, dimmap=dimmap)
+        _broadcast!((x,y)->f(y,x), out, B, A, dimmap=dimmap)
     else
-        _broadcast!(f, similar(A), A, B, dimmap=dimmap)
+        _broadcast!(f, out, A, B, dimmap=dimmap)
     end
 end
 
--- a/src/join.jl
+++ b/src/join.jl
@@ -424,7 +424,7 @@ If `dimmap` is not specified, it is determined automatically using index column
 names and types.
 """
 function broadcast(f::Function, A::IndexedTable, B::IndexedTable; dimmap=nothing)
-    out_T = Base.promote_op(f, eltype(A), eltype(B))
+    out_T = _promote_op(f, eltype(A), eltype(B))
     if ndims(B) > ndims(A)
         out = IndexedTable(similar(B.index, 0), similar(arrayof(out_T), 0))
         _broadcast!((x,y)->f(y,x), out, B, A, dimmap=dimmap)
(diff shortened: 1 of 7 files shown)


simonbyrne:
Thanks, but it seems to break when one of the index columns is a `PooledArray`.
```
julia> typeof(A.index)
IndexedTables.Columns{Tuple{String,Int64,Int64},NamedTuples._NT_Class_Scen_Year{PooledArrays.PooledArray{String,UInt8,1,Array{UInt8,1}},Array{Int64,1},Array{Int64,1}}}

julia> similar(typeof(A.index),0)
ERROR: MethodError: Cannot `convert` an object of type Tuple{Int64} to an object of type PooledArrays.PooledArray{String,UInt8,1,Array{UInt8,1}}
This may have arisen from a call to the constructor PooledArrays.PooledArray{String,UInt8,1,Array{UInt8,1}}(...),
since type constructors fall back to convert methods.
Stacktrace:
 [1] PooledArrays.PooledArray{String,UInt8,1,Array{UInt8,1}}(::Tuple{Int64}) at ./sysimg.jl:24
 [2] similar(::Type{T} where T, ::Int64) at ./abstractarray.jl:565
 [3] map(::IndexedTables.##28#29{Int64}, ::SimpleVector) at ./essentials.jl:274
 [4] similar(::Type{IndexedTables.Columns{Tuple{String,Int64,Int64},NamedTuples._NT_Class_Scen_Year{PooledArrays.PooledArray{String,UInt8,1,Array{UInt8,1}},Array{Int64,1},Array{Int64,1}}}}, ::Int64) at /Users/simon/.julia/v0.6/IndexedTables/src/columns.jl:52
```

src/join.jl
```diff
@@ -424,10 +424,13 @@ If `dimmap` is not specified, it is determined automatically using index column
 names and types.
 """
 function broadcast(f::Function, A::IndexedTable, B::IndexedTable; dimmap=nothing)
+    out_T = Base.promote_op(f, eltype(A), eltype(B))
+    out = IndexedTable(similar(typeof(A.index), 0), similar(arrayof(out_T), 0))
```
simonbyrne: Seems to work if we replace `similar(typeof(A.index), 0)` with `similar(A.index, 0)`, though shouldn't this go inside the if statement below (because in the second case, you would want `B` instead of `A`?

simonbyrne:
Also, what should be the type of a named column? e.g. I get:
```
julia> out_T = Base.promote_op(f, eltype(A), eltype(B))
NamedTuples._NT_CorrScen{Int64}

julia> IndexedTables.arrayof(out_T)
IndexedTables.Columns{NamedTuples._NT_CorrScen{Int64},Tuple{Array{Int64,1}}}
```
Shouldn't that last line be 
```
IndexedTables.Columns{NamedTuples._NT_CorrScen{Int64},NamedTuples._NT_CorrScen{Array{Int64,1}}}
```
?

simonbyrne:
Okay, I've pushed a patch that should fix both of those. Note the change to `arrayof`: I'm not sure if the `eval` will cause problems.

src/utils.jl
```diff
@@ -279,3 +280,43 @@ Base.@pure function arrayof(T)
...
+@inline function _promote_op{S}(f, ::Type{S})
+    t = Core.Inference.return_type(f, Tuple{Base._default_type(S)})
+    strip_unionall(t)
+end
+
+@inline function _promote_op{S,T}(f, ::Type{S}, ::Type{T})
+    t = Core.Inference.return_type(f, Tuple{Base._default_type(S),
+                                        Base._default_type(T)})
+    strip_unionall(t)
+end
```
shashi: @JeffBezanson Could Base use some form of this? It allows more information to propagate when the return type is a Tuple where some elements are concrete and some are Any/UnionAll.

Also, what is the reason to remove `UnionAll`s in `Base.promote_op` as it does now?

PR #62, opened 2017-07-17
Don't force index Columns eltype to be Tuple

shashi:
Scalar indexing now converts the tuple to NamedTuple and does the lookup.

### DIFF
--- a/src/IndexedTables.jl
+++ b/src/IndexedTables.jl
@@ -17,11 +17,11 @@ const DimName = Union{Int,Symbol}
 include("utils.jl")
 include("columns.jl")
 
-immutable IndexedTable{T, D<:Tuple, C<:Tup, V<:AbstractVector}
-    index::Columns{D,C}
+immutable IndexedTable{T, D<:Tuple, C<:Columns, V<:AbstractVector}
+    index::C
     data::V
 
-    index_buffer::Columns{D,C}
+    index_buffer::C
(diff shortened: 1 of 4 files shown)


src/IndexedTables.jl
```diff
@@ -17,11 +17,11 @@ const DimName = Union{Int,Symbol}
 include("utils.jl")
 include("columns.jl")
 
-immutable IndexedTable{T, D<:Tuple, C<:Tup, V<:AbstractVector}
-    index::Columns{D,C}
+immutable IndexedTable{T, D<:Tuple, C<:Columns, V<:AbstractVector}
```
JeffBezanson: Maybe we don't need the `D` parameter at all now?

src/indexing.jl
```diff
@@ -54,7 +54,7 @@ end
 function _getindex(t::IndexedTable, idxs)
     I = t.index
     cs = astuple(I.columns)
-    if length(idxs) != length(I.columns)
+    if nfields(idxs) !== nfields(I.columns)
```
JeffBezanson: Doesn't `length` work on both? Better to use that since it's a more general function.

src/IndexedTables.jl
shashi: Because we have `eltypes`? You maybe right, it's seldom used.

Issue #63, opened 2017-07-17
RFC: Iteration API


shashi:
My current thinking for a coherent API:

1. `columns` -- returns (named) tuples of column vectors themselves.
```julia
columns(c::Columns) = c.columns
columns(c::Columns, which...) = c.columns[[which...]]
columns(t::IndexedTable) = (t.index.columns..., t.data.columns...)
  # ^ named tuple if everything is named
columns(t::IndexedTable, which...) = all columns with those names
  # ^ this one assumes uniqueness of names / continuity of numbering
  # among index and data cols
```
Not sure how useful this is for users, but it's useful to implement row iterators below.

2. `rows` -- iterator over rows.

```julia
rows(c::Columns) = c
rows(c::Columns, which...) = Columns(columns(c, which...))
rows(t::IndexedTable) = Columns(columns(t))
rows(c::IndexedTable, which...) = Columns(columns(t, which...))
```

It's a little funny that `columns` returns a tuple, and `rows` returns `Columns`. :) I find this quite apt, but will admit it might be confusing because of the naming.

3. `keys`, `values` and `pairs` -- iterators over rows that acknowledge the key-value pair view of IndexedTable
```julia
keys(t::IndexedTable, dims...) = Columns(columns(keys(t), dims...))
values(t::IndexedTable, dims...) = Columns(columns(values(t), dims...))

pairs(t::IndexedTable) = Columns(keys(t), values(t)) # nested Columns
```
Bonus feature:

`which` arguments to `columns` (and hence `dims` args in other operations) can be `Symbol`, `Int` or *`As`* --[`As(f, src_name, dest_name)`](https://github.com/JuliaComputing/IndexedTables.jl/blob/37bb28e8f7e57fd874982af5454fd4cc11ad4ff2/src/query.jl#L33-L39) will cause a field to be renamed in the output named tuple and optionally apply a function `f` on it before jamming it into the output.

The exciting part is, all these can be implemented on `DTable`. If we had a common name for an IndexedTable constructor, (say `table(keys, vals)`), then we can dispatch to create either a `DTable` or an `IndexedTable` based on whether the inputs are distributed or not. I can imagine some operations (e.g. `select`, `permutedims`) could have a common implementation.

I will rebase #55 against this.

cc @JeffBezanson @StefanKarpinski @davidanthoff

JeffBezanson:
Looks good to me!

[...]

PR #64, opened 2017-07-18
WIP/RFC: Iteration API

shashi:
See #63 

One minor annoyance:

`keys(t::IndexedTable, col::Union{Symbol,Int})` here returns a `Columns` with 1 vector. To get the column, you'd have to do `column(keys(t), col)`... This is so that `keys(t, cols...)`  always works by returning `Columns`. There's an alternative where you could make `keys(t, col::Tuple)` return columns but `keys(t, col::Symbol)` return the vector and not have `keys(t, cols...)`. Not sure which one to go with. @JeffBezanson what do you think?

### DIFF
--- a/src/IndexedTables.jl
+++ b/src/IndexedTables.jl
@@ -215,6 +215,12 @@ identified by `which`.
 """
 rows(t::IndexedTable, which...) = rows(columns(t, which...))
 rows(t::Columns, which...) = rows(columns(t, which...))
+function rows(t::AbstractVector, which...)
+    if all(x->x==1, which)
+        Columns(map(x->t, which))
+    else error("No column $(join(filter(x->x!=1, which), " "))")
+    end
+end
 
 ## Row-wise iteration that acknowledges key-value nature
 
@@ -236,7 +242,7 @@ keys(t::IndexedTable, which...) = rows(keys(t), which...)
 """
 `values(t)`
 
-Returns an array of values stored `t`.
+Returns an array of values stored in `t`.
 """
 values(t::IndexedTable) = t.data
 
@@ -247,13 +253,6 @@ Returns a array of rows from a subset of columns
 of the values in `t`.
 """
 values(t::IndexedTable, which...) = rows(values(t), which...)
-
-"""
-`pairs(t)`
-
-Returns an array of key-value pairs in `t`
-"""
-pairs(t::IndexedTable) = Columns(keys(t), values(t))
 
 ## As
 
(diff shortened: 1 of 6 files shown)


src/IndexedTables.jl
```diff
@@ -213,17 +382,17 @@ map{T,D<:Tuple,C<:Tup,V<:Columns}(p::Proj, x::IndexedTable{T,D,C,V}) =
...
+# Given an IndexedTable array with multiple data columns (its data vector is a `Columns` object), return a
+# new array with the specified subset of data columns. Data is shared with the original array.
+# """
+# columns(x::IndexedTable, which...) = IndexedTable(x.index, Columns(x.data.columns[[which...]]), presorted=true)
 
-columns(x::IndexedTable, which) = IndexedTable(x.index, x.data.columns[which], presorted=true)
+#columns(x::IndexedTable, which) = IndexedTable(x.index, x.data.columns[which], presorted=true)
 
-column(x::IndexedTable, which) = columns(x, which)
+#column(x::IndexedTable, which) = columns(x, which)
```
shashi: Not sure how to deprecate this...

src/utils.jl
```diff
@@ -358,3 +370,13 @@ Base.@pure @generated function map_params{T<:NamedTuple,S<:NamedTuple}(f, ::Type
     NT = Expr(:macrocall, :(NamedTuples.$(Symbol("@NT"))), fieldnames(T)...)
     :($NT{_map_params(f, T, S)...})
 end
+
+Base.@pure function concat_nt_type{
+           T<:NamedTuple,S<:NamedTuple}(::Type{T}, ::Type{S})
+    namedtuple(fieldnames(T)..., fieldnames(S)...)
+end
```
shashi: The `@pure` here is unsuccessful at allowing inference for this.

shashi:
I kept the varargs API for getting subsets of columns. I think this PR is in its final form...

shashi:
To make it easier to comment on, here's the change resulted by the alternative API:

https://github.com/JuliaComputing/IndexedTables.jl/commit/e57f0e78a104815d67bd11ef7a3d1850826a1efc

Some awkward things about this:

- `columns(t, :col)` -- is a Vector (the column)
- `rows(t, :col)` -- is also a Vector and is the same as `column(t, :col)`
- `rows(t, :col1, :col2)` is now `rows(t, (:col1, :col2))`
- `rows(t, as(2,:x))` -- gives the vector, so the column has no name
- `rows(t, (as(2, :x),))` -- gives the Columns with the name "x"

But the upside is:

- `keys(t, :col)` and `values(t, :col)` are a Vectors

I'm still ambivalent about which one is good.

tkelman:
`as` and `As` could use docs, not super clear what those are

PR #76, opened 2017-08-23
Add TableTraits.jl integration

davidanthoff:


JeffBezanson:
Cool, thanks David! Will there eventually be a way to get a length (or estimate) from the iterator, or better yet reuse whole columns in place?

src/tabletraits.jl
```diff
@@ -0,0 +1,72 @@
...
+end
+
+function IndexedTable(x; idxcols::Union{Void,Vector{Symbol}}=nothing, datacols::Union{Void,Vector{Symbol}}=nothing)
+    if isiterabletable(x)
+        iter = getiterator(x)
+
+        source_colnames = TableTraits.column_names(iter)
+        source_coltypes = TableTraits.column_types(iter)
+
+        if idxcols==nothing && datacols==nothing
```
shashi: It seems a natural conversion of `Columns` to `IndexedTable` is to have a `1:n` index: it's the same as a 1-d array of named tuples.

src/tabletraits.jl
```diff
@@ -0,0 +1,72 @@
...
+    end
+
+    quote
+        for row in iter
+            $push_exprs
+        end
+    end
+end
+
+function IndexedTable(x; idxcols::Union{Void,Vector{Symbol}}=nothing, datacols::Union{Void,Vector{Symbol}}=nothing)
```
davidanthoff: Yes, you should just be able to add another method that handles that case, right? It would be good if the named arguments had the same semantics, of course.

I'm also not sure this is the right API, I just was loosely inspired by [this](https://github.com/JuliaComputing/NDSparseIO.jl/blob/master/src/NDSparseIO.jl#L29).

src/tabletraits.jl
davidanthoff: Index columns need unique values, right? So one could also say that in this case it should create an ``IndexedTable`` without any index, so that this conversion works for any table data without having things like a unique requirement. But I'm not sure, I think the main thing is to be consistent across the different ways to create things both in ``IndexedTable`` and ``JuliaDB``.

src/tabletraits.jl
davidanthoff: Oh, one more thing: we should also add a code path to this method that deals with an iterator where the element type is ``Pair{X,S}``. If it is just any ``Pair``, it would create an unnamed index and data column. If either ``X`` or ``X`` are a ``NamedTuple``, it would create named columns for the index and data columns. At that point the following would automatically work:
````julia
@from i in source begin
    @select {i.a, i.b} => {i.c,i.d}
    @collect IndexedTable
end
````
Not in this PR, but could be added later.

davidanthoff:
> Will there eventually be a way to get a length (or estimate) from the iterator

Yes, the whole design here just follows the iterator interface in base. So if the source returns ``HasLength()`` from ``iteratorsize``, ``length`` will work. In fact, I think almost all the table sources support that, and in Query.jl queries the ``HasLength()`` property also gets preserved if possible (i.e. operations like ``@select`` preserve it, ``@filter`` obviously doesn't).

So the issue here is just that this PR is not the most efficient implementation :) AND I just realized that this PR is a bit silly. https://github.com/davidanthoff/TableTraitsUtils.jl has a default implementation for the iterable tables trait that handles all these things properly, and there is actually no reason why I couldn't just use that implementation here as well. I just overlooked this, the code in this PR predates those efficient implementations and I forgot to update things. I'll update this PR tomorrow or so to use all of that.

One thing that would help even more is JuliaLang/julia#22467, in that case even more Query.jl queries could use preallocated arrays of the right size when they get materialized. I had a very hackish version of something like that and it allowed me to get my ``@groupby`` implementation to within a factor of 2x of the pandas code in terms of speed.

> , or better yet reuse whole columns in place?

Not yet, but I'm mulling some designs for that. I think that can come later, though, I would add it as an optional, more performant way to the table traits interface.

davidanthoff:
Alright, the new commit now uses the [TableTraitsUtils.jl](https://github.com/davidanthoff/TableTraitsUtils.jl) implementation to generate the vectors that hold the data, and that implementation uses an optimized [codepath](https://github.com/davidanthoff/TableTraitsUtils.jl/blob/master/src/TableTraitsUtils.jl#L130) for sources that return ``Base.HasLength()``.

shashi:
LGTM.

davidanthoff:
Cool.

What is your plan for tagging a new version? I should try to release a version of IterableTables.jl (that no longer has the integration code in it) ideally more or less simultaneously so that folks don't get override warnings.

shashi:
The package is ready to be tagged. Just ping me on a metadata PR and I'll
make one against it.

PR #79, opened 2017-09-03
WIP: `by` and `with`

shashi:
1) Adds `sortperm` on IndexedTable by a subset of columns which uses the existing sortedness of indexed columns.
2) `sortproxy` - makes a UInt<smallest> identically sorted label array for any array (O(n)) -- can then be used for `O(n)` sort/perm
2) Adds ability to aggregate and perform joins in a custom permute order
3) Uses 1) & 2) to allow aggregation by combinations of indexed and non-indexed columns (kinda) efficiently. Any improvements to `sortperm` / `refine_perm!` should improve aggregation performance.

API changes:
- `aggregate`, `aggregate_vec` and `aggregate!` take `by` and `with` keyword arguments (column name / number or tuple of column names / numbers). Aggregation is done by `by` columns, reduction is done with the `with` column(s).
- `select(arr, conditions::Pair...)` can now take Pairs of column name => functions where the column name can be both key or value columns
- `reducedim()` takes a `with` kwarg which is passed to f. (also deleted `_aggregate_vec!` no longer needed here)
- `innerjoin`, `leftjoin` and `asofjoin` take `by`, `with`, `leftby`, `rightby`, `leftwith`, `rightwith`

TODO:
- [ ] benchmark the joins
- [ ] implement options for `merge`
- [ ] tests

### DIFF
--- a/src/IndexedTables.jl
+++ b/src/IndexedTables.jl
@@ -169,7 +169,7 @@ function column(t::IndexedTable, col::Symbol)
         return column(values(t), col)
     end
 
-    error("Couldn't find column named $n")
+    error("Couldn't find column named $col")
 end
 
 ## Column-wise iteration:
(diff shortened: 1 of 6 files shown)


shashi:
I've merged most of this piece by piece. The rest is in #85

Issue #82, opened 2017-09-26
Major API changes


shashi:
See https://github.com/JuliaComputing/JuliaDB.jl/issues/80 for more. I'd like any discussion to be consolidated there.

PR #85, opened 2017-10-13
WIP: API changes

shashi:
Starting off #82 

- [x] Deprecate `IndexedTable` for `NDSparse`
- [x] Introduce `NextTable`
- [x] `select`, `filter`, `groupreduce`, `groupby`
- [x] innerjoin
- [x] outerjoin
- [x] leftouterjoin
- [ ] windowed/asofjoin
- [x] infer output column names


shashi:
API question:

Should join functions by default join on:
1. Columns that match in name OR
2. Primary index columns

Should join functions by default join with: (i.e. pass as argument to `f`)

1. All fields but those that are being joined "ON"
2. All fields (entire row)

SQL prefers 1) and 1), pandas uses 2) and 1)

This PR uses 2) and 2). I think I'd prefer the SQL route for the Table datastructure, and the pandas route for `NDSparse`.

simonbyrne:
Not sure what you mean: can you give examples of each case?

shashi:
Say we have two tables
```
A: room_id, start_time, program
```
Indexed by `room_id, start_time`

and
```
B: room_id,floor, area
```
indexed by `room_id, floor`

when we do `join(A, B)` by default in case

1. We'd join by room_id
2. We'd join by `room_id, start_date` = `room_id, floor_no` --> this is obviously wrong and might fail

the two cases of how the columns are combined would be, the function `f` passed to join would get:

1. (`@NT(start_time, program)`, `@NT(floor_no, area)`) (if using 1) above), or (`@NT(program)`, `@NT(area)`) if using 2) above.
2. (`@NT(room_id, start_time, program)`, `@NT(room_id, floor_no, area)`)

andyferris:
I'd say 1) and 2)?

I feel the default join should be the natural join. These semantics make sense for other table types that might not have acceleration indices at all, but joining still makes sense.

It's a very interesting question whether you want the joined column(s) returned. I'd suggest to keep them by default, since It is pretty natural to assume that the joined column(s) provides a good index for the output table (this should also be true of other table packages: a hash map or some such will be constructed/used for the join, and it seems a shame to discard it).

(BTW, [looking at pandas](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.join.html) it seems the situation isn't very clear - either the join column is kept as data or as an index, so I think the analogy here is to keep that column as the `IndexedTable`'s sorted index column)

PR #86, opened 2017-10-15
Updated readme about multiple value columns constructor

xiaodaigh:


shashi:
Much appreciated! :+1:

PR #88, opened 2017-10-18
Add empty dst check in _merge!

getzdan:
When merging an IndexedTable into an empty one, we can just copy. Without this short-circuit check an invalid indexing is used.


shashi:
Thanks! :+1:

Issue #92, opened 2017-11-23
IndexedTable from TextParse.csvread output


tcovert:
I couldn't find an example in the documentation for how one builds an `IndexedTable`, with column names, from the output of `TextParse.csvread`.  This appears to work (took me a while...):

```
tp = csvread("file.csv")
t = table(tp[1]..., names = map(Symbol, tp[2]))
```

I would have thought something like `table(csvread("file.csv"))` or even `table(Columns(csvread("file.csv")))` would work but both give an error like:
```
julia> table(TextParse.csvread("file.csv", escapechar='"', nastrings = vcat(TextParse.NA_STRINGS, "Confidential")))
ERROR: MethodError: no method matching _impl(::Tuple{Array{Int64,1},Array{Int64,1},Array{String,1},Array{String,1},Array{String,1},Array{String,1},Array{String,1},Array{String,1},DataValues.DataValueArray{Date,1},DataValues.DataValueArray{Int64,1},Array{String,1},Array{String,1},Array{String,1},Array{Int64,1},PooledArrays.PooledArray{String,UInt8,1,Array{UInt8,1}},Array{String,1},Array{String,1},Array{String,1},Array{String,1},Array{String,1},PooledArrays.PooledArray{String,UInt8,1,Array{UInt8,1}},DataValues.DataValueArray{Float64,1},DataValues.DataValueArray{Float64,1},PooledArrays.PooledArray{String,UInt8,1,Array{UInt8,1}},PooledArrays.PooledArray{String,UInt8,1,Array{UInt8,1}},DataValues.DataValueArray{Int64,1},DataValues.DataValueArray{Date,1}}, ::Array{String,1})
``` 

I guess this isn't a bug, but if the goal is for the first syntax above to be the recommended syntax, it might be helpful for there to be an example in the docstrings...

davidanthoff:
I'll fix the iterable tables integration story for the new style API here. Then this
````julia
load("file.scv") |> table
# or
table(load("file.csv"))
````
should work.

I've also been sitting on a small extension to TableTraits.jl that will remove the perf overhead of going through the iterable tables system, i.e. at that point it should give the raw performance of TextParse (which is used by the FileIO story under the hood).

davidanthoff:
Ha, I just realized that this new ``NextTable`` is already an iterable table source, without any integration code, simply because it iterates named tuples! Very nice. So all of the following already works:
````julia
t = table(...)

# File IO
save("file.csv", t)
[...]

PR #93, opened 2017-11-23
Update tabletraits

davidanthoff:
This should address #92.

### DIFF
--- a/src/tabletraits.jl
+++ b/src/tabletraits.jl
@@ -4,7 +4,7 @@ using TableTraitsUtils
 TableTraits.isiterable(x::NDSparse) = true
 TableTraits.isiterabletable(x::NDSparse) = true
 
-function TableTraits.getiterator{S<:NDSparse}(source::S)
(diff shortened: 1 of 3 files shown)


src/tabletraits.jl
```diff
@@ -46,3 +46,25 @@ function NDSparse(x; idxcols::Union{Void,Vector{Symbol}}=nothing, datacols::Unio
...
 end
+
+function table(rows::AbstractArray{T}; kwargs...) where {T<:NamedTuple}
+    source_data, source_names = TableTraitsUtils.create_columns_from_iterabletable(rows)
+    
+    kwargs_dict = Dict(i[1]=>i[2] for i in kwargs)   
+    kwargs_dict[:copy] = false
+
+    return table(source_data..., names=source_names; kwargs_dict...)
+end
```
shashi: This is ambiguous with the constructor for a table with 1 column which are named tuples. Normally I'm not a big fan of special cases like this, but this one might just be OK.

PR #94, opened 2017-11-23
Add toplevel testset

davidanthoff:
I like that the test results all appear in one table with this. Purely aesthetic, so if you like it some other way, just close this :)


shashi:
Thanks!

Issue #105, opened 2017-12-07
Can't generate tables from Columns of not Arrays


piever:
I'm not sure if this is intended. I can create a Columns object even if the arguments are not strictly arrays, but then I cannot convert it to table:

```julia
julia> c = Columns(1:10, rand(10))
10-element Columns{Tuple{Int64,Float64}}:
 (1, 0.187685) 
 (2, 0.364163) 
 (3, 0.650937) 
 (4, 0.646284) 
 (5, 0.995686) 
 (6, 0.700621) 
 (7, 0.217061) 
 (8, 0.935344) 
 (9, 0.859523) 
 (10, 0.414292)

julia> table(c)
ERROR: MethodError: Cannot `convert` an object of type Array{Int64,1} to an object of type UnitRange{Int64}
This may have arisen from a call to the constructor UnitRange{Int64}(...),
since type constructors fall back to convert methods.
Stacktrace:
 [1] IndexedTables.Columns{Tuple{Int64,Float64},Tuple{UnitRange{Int64},Array{Float64,1}}}(::Tuple{Array{Int64,1},Array{Float64,1}}) at /Users/pietro/.julia/v0.6/IndexedTables/src/columns.jl:25
 [2] similar(::IndexedTables.Columns{Tuple{Int64,Float64},Tuple{UnitRange{Int64},Array{Float64,1}}}) at /Users/pietro/.julia/v0.6/IndexedTables/src/columns.jl:189
 [3] copy(::IndexedTables.Columns{Tuple{Int64,Float64},Tuple{UnitRange{Int64},Array{Float64,1}}}) at ./abstractarray.jl:698
 [4] #table#91(::Array{Int64,1}, ::Void, ::Array{IndexedTables.Perm,1}, ::Bool, ::Bool, ::Array{Nullable{Float64},1}, ::Function, ::Val{:serial}, ::Tuple{UnitRange{Int64},Array{Float64,1}}) at /Users/pietro/.julia/v0.6/IndexedTables/src/table.jl:214
 [5] (::IndexedTables.#kw##table)(::Array{Any,1}, ::IndexedTables.#table, ::Val{:serial}, ::Tuple{UnitRange{Int64},Array{Float64,1}}) at ./<missing>:0
 [6] #table#95(::Void, ::Array{Any,1}, ::Function, ::Tuple{UnitRange{Int64},Array{Float64,1}}) at /Users/pietro/.julia/v0.6/IndexedTables/src/table.jl:247
[...]

PR #106, opened 2017-12-07
WIP: Added view of NextTable

piever:
Work in progress implementation of a view of NextTable. I still have a couple of doubts that I'll leave as comments to the code.

### DIFF
--- a/src/columns.jl
+++ b/src/columns.jl
@@ -2,7 +2,7 @@ using Compat
 
 import Base:
     linearindexing, push!, size, sort, sort!, permute!, issorted, sortperm,
-    summary, resize!, vcat, serialize, deserialize, append!, copy!
+    summary, resize!, vcat, serialize, deserialize, append!, copy!, view
 
 export Columns, colnames, ncols, ColDict, insertafter!, insertbefore!, @cols, setcol, pushcol, popcol, insertcol, insertcolafter, insertcolbefore, renamecol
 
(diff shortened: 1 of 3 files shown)


src/columns.jl
```diff
@@ -186,8 +186,17 @@ Base.IndexStyle(::Type{<:Columns}) = IndexLinear()
 summary(c::Columns{D}) where {D<:Tuple} = "$(length(c))-element Columns{$D}"
 
 empty!(c::Columns) = (foreach(empty!, c.columns); c)
-similar(c::Columns{D,C}) where {D,C} = Columns{D,C}(map(similar, c.columns))
-similar(c::Columns{D,C}, n::Integer) where {D,C} = Columns{D,C}(map(a->similar(a,n), c.columns))
+
+function similar(c::Columns{D,C}) where {D,C}
+    cols = map(similar, c.columns)
+    Columns{D,typeof(cols)}(cols)
```
piever: How bad is it to lessen the inferrability of the output?

src/columns.jl
```diff
@@ -210,6 +219,8 @@ getindex(c::Columns{D}, i::Integer) where {D<:NamedTuple} = D(ith_all(i, c.colum
 
 getindex(c::Columns, p::AbstractVector) = Columns(map(c->c[p], c.columns))
 
+view(c::Columns, I) = Columns(map(a->view(a, I), c.columns))
+
```
piever: Or:
```
function view(c::Columns{D,C}, I) where {D,C} 
     cols = map(a->view(a, I), c.columns)
     Columns{D,typeof(cols)}(cols)
 end
```
Not sure what are the implication of the explicit typing.

src/table.jl
```diff
@@ -289,6 +289,16 @@ Base.getindex(t::NextTable, i::Integer) = getindex(t.columns, i)
 Base.getindex(t::NextTable, i::Colon) = copy(t)
 Base.endof(t::NextTable) = length(t)
 
+function Base.view(t::NextTable, I)
+    issorted(I) || throw(ArgumentError("`view` called with unsorted index."))
+    NextTable(
+        view(t.columns, I),
+        t.pkey,
+        Perm[],
+        fill(Nullable{Float64}(), length(t.columns)),
```
piever: I think it's important to keep the primary keys, but I'm not sure what `Perm` does and whether it can be kept.

src/columns.jl
shashi: This call should be inferrable. :+1:

src/columns.jl
shashi: They should be equivalent.

As long as `view` on the columns are inferrable, this will work.

src/table.jl
shashi: You can throw away `Perm`s as you wish. It's not exposed functionality right now anyway.

shashi: `Perm` stores a permutation (output of `sortpermby`) and which columns this sortperm was performed by. So if you're repeatedly grouping or joining by the same subset of columns you can use `join(x,y, cache=true)` and have that be cached...

piever:
I've incorporated feedback from #105 and added tests: it's good to go on my side.

shashi:
Excellent!

PR #107, opened 2017-12-12
WIP: Readme

joshday:
The readme is a bit out of date.  This PR does some cleanup, removes most of the content, and instead points people towards the JuliaDB API reference.  Thoughts?


joshday:
@shashi please correct me if the claims about performance between table types is incorrect.

PR #108, opened 2017-12-16
Use DataValueArray for TableTraits integration

davidanthoff:
This will use ``DataValueArray`` for columns that have missing data during construction via the iterable tables interface. Currently one ends up with ``Vector{DataValue}``, which seems not ideal. I was originally hesitant to use ``DataValueArray`` here because I didn't want to "shove" it onto IterableTables.jl, but it seems that you guys are using it in any case already, so this might make sense?

### DIFF
--- a/src/tabletraits.jl
+++ b/src/tabletraits.jl
@@ -8,6 +8,16 @@ function TableTraits.getiterator(source::S) where {S <: NDSparse}
     return rows(source)
 end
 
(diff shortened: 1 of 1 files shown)


davidanthoff:
Bump, let me know if you think there is anything else I should do here.

davidanthoff:
@shashi Bump again, this would make things a lot more efficient.

PR #109, opened 2018-01-03
WIP: added onall to summarize all columns

piever:
This adds 3 special selectors. `OnAll` to apply a list of functions to all columns, `Not` for the complementary selection and `Keys` to select the primary keys. This allows:

```julia
Main> using JuliaDB, IndexedTables

Main> iris = loadtable("/Users/pietro/downloads/iris.csv")
Table with 150 rows, 5 columns:
SepalLength  SepalWidth  PetalLength  PetalWidth  Species
─────────────────────────────────────────────────────────────
5.1          3.5         1.4          0.2         "setosa"
4.9          3.0         1.4          0.2         "setosa"
4.7          3.2         1.3          0.2         "setosa"
4.6          3.1         1.5          0.2         "setosa"
⋮
6.5          3.0         5.2          2.0         "virginica"
6.2          3.4         5.4          2.3         "virginica"
5.9          3.0         5.1          1.8         "virginica"

Main> groupby(OnAll(mean, std), iris, :Species, select = Not(:Species))

Table with 3 rows, 9 columns:
Species       SepalLength_mean  SepalWidth_mean  PetalLength_mean  PetalWidth_mean  SepalLength_std  SepalWidth_std  PetalLength_std  PetalWidth_std
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
"setosa"      5.006             3.418            1.464             0.244            0.35249          0.381024        0.173511         0.10721
"versicolor"  5.936             2.77             4.26              1.326            0.516171         0.313798        0.469911         0.197753
"virginica"   6.588             2.974            5.552             2.026            0.63588          0.322497        0.551895         0.27465
```

I still have to add docstrings and tests. I have a very strange bug in the REPL though (the above works in the VS Code extension). In the REPL with the exact same code I get:

```julia
julia> groupby(OnAll(mean, std), iris, :Species, select = Not(:Species))
Table with 3 rows, 9 columns:
Columns:
#  colname           type
────────────────────────────
1  Species           String
2  SepalLength_mean  Float64
3  SepalWidth_mean   Float64
4  PetalLength_mean  Float64
5  PetalWidth_mean   Float64
6  SepalLength_std   Float64
7  SepalWidth_std    Float64
8  PetalLength_std   Float64
9  PetalWidth_std    Float64
```

which really doesn't make a lot of sense to me. Any idea what is going on?

### DIFF
--- a/src/selection.jl
+++ b/src/selection.jl
@@ -1,4 +1,4 @@
-export dropna, selectkeys, selectvalues
+export dropna, selectkeys, selectvalues, ApplyColwise, Not, Keys
 
 """
 `select(t::Table, which::Selection)`
@@ -142,6 +142,39 @@ end
 
 function selectvalues(x::NDSparse, which; kwargs...)
     ndsparse(keys(x), rows(values(x), which); kwargs...)
+end
+
+struct ApplyColwise{N, T<:Tuple}
+    functions::T
+    names::NTuple{N, Symbol}
+end
+
+ApplyColwise(args...) = ApplyColwise(args)
+ApplyColwise(t::Tuple) = ApplyColwise(t, map(Symbol,t))
+
+function (ac::ApplyColwise)(t::Union{AbstractIndexedTable, Columns})
+    func = Tuple(Symbol(s, :_, n) => s => f for s in colnames(t),
+        (f, n) in zip(ac.functions, ac.names))
+    fs, input, S = init_inputs(func, t, reduced_type, true)
+    _apply(fs, fs isa Tup ? columns(input) : input)
+end
+
+(ac::ApplyColwise)(t) = (ac::ApplyColwise)(convert(Columns, t))
+
+struct Not{T}
+    names::T
+end
+
+Not(args...) = Not(args)
+
+for f in [:(Base.select), :columns, :rows, :excludecols]
+    @eval $f(t::AbstractIndexedTable, not::Not) = $f(t, excludecols(t, not.names))
+end
+
+struct Keys; end
+
+for f in [:(Base.select), :columns, :rows, :excludecols]
+    @eval $f(t::AbstractIndexedTable, keys::Keys) = $f(t, pkeynames(t))
 end
 
 """
--- a/src/selection.jl
+++ b/src/selection.jl
@@ -1,4 +1,4 @@
-export dropna, selectkeys, selectvalues, ApplyColwise, Not, Keys
+export dropna, selectkeys, selectvalues, Not, Keys
 
 """
 `select(t::Table, which::Selection)`
@@ -143,23 +143,6 @@ end
 function selectvalues(x::NDSparse, which; kwargs...)
     ndsparse(keys(x), rows(values(x), which); kwargs...)
 end
-
-struct ApplyColwise{N, T<:Tuple}
-    functions::T
-    names::NTuple{N, Symbol}
-end
-
-ApplyColwise(args...) = ApplyColwise(args)
-ApplyColwise(t::Tuple) = ApplyColwise(t, map(Symbol,t))
-
-function (ac::ApplyColwise)(t::Union{AbstractIndexedTable, Columns})
-    func = Tuple(Symbol(s, :_, n) => s => f for s in colnames(t),
-        (f, n) in zip(ac.functions, ac.names))
-    fs, input, S = init_inputs(func, t, reduced_type, true)
-    _apply(fs, fs isa Tup ? columns(input) : input)
-end
-
-(ac::ApplyColwise)(t) = (ac::ApplyColwise)(convert(Columns, t))
 
 struct Not{T}
     names::T
--- a/src/reduce.jl
+++ b/src/reduce.jl
@@ -413,18 +413,21 @@ end
 ApplyColwise(args...) = ApplyColwise(args)
 ApplyColwise(t::Tuple) = ApplyColwise(t, map(Symbol,t))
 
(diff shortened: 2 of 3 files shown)


piever:
Actually the result was correct also in the REPL (it would really have been too weird otherwise), it simply gets displayed differently, sorry about the noise... Is it intended to change the display according to the number of columns?

piever:
I've changed slightly `groupby` to allow the case where there are no grouping keys (`by == ()`), as it is actually useful in combination with `OnAll` when one wants to summarize columns (here `by == ()` by default as `select(iris, Not(:Species))` has no primary columns):

```julia
julia> groupby(OnAll(mean,std), select(iris, Not(:Species)))
(SepalLength_mean = 5.843333333333334, SepalWidth_mean = 3.054, PetalLength_mean = 3.758666666666666, PetalWidth_mean = 1.1986666666666665, SepalLength_std = 0.828066127977863, SepalWidth_std = 0.4335943113621737, PetalLength_std = 1.7644204199522626, PetalWidth_std = 0.7631607417008412)
```

Maybe we want to find another name for this operation though.

shashi:
You can disable the compact display using: `IndexedTables.set_show_compact!(false)` it swiches to compact if there isn't enough horizontal space for a row.

shashi:
`groupby(OnAll(mean,std), select(iris, Not(:Species)))`

To me it seems like this should group by all but the species column. - Not what we want.

I think the correct semantics here would be

```julia
julia> OnAll(mean, std)(select(table, Not(:Species)))
(SepalLength_mean = 5.843333333333334, SepalWidth_mean = 3.054, PetalLength_mean = 3.758666666666666, PetalWidth_mean = 1.1986666666666665, SepalLength_std = 0.828066127977863, SepalWidth_std = 0.4335943113621737, PetalLength_std = 1.7644204199522626, PetalWidth_std = 0.7631607417008412)

```
A `(::OnAll)(x)`
method would naturally make its use in `groupby` also work.

src/selection.jl
```diff
@@ -144,6 +144,34 @@ function selectvalues(x::NDSparse, which; kwargs...)
...
+    (f, n) in zip(onall.functions, onall.names))
+
+struct Not{T}
+    names::T
+end
+
+Not(args...) = Not(args)
+
+for f in [:(Base.select), :columns, :rows, :excludecols]
+    @eval $f(t::AbstractIndexedTable, not::Not) = $f(t, excludecols(t, not.names))
```
shashi: Just defining `columns(x, ::Not)` should make `rows` work.

shashi: I tried something like this a few days ago, I think the right way to do this would be to have something like this:

```julia
column(x, y::Union{Not, Keys}) = rows(x, lowerselection(x, y))
```
Where `lowerselection(x, y)` returns a tuple of column numbers...

This definition also answers this question: given
```julia
julia> t = table([1], [2], [3], names=[:foo,:bar,:baz])
Table with 1 rows, 3 columns:
foo  bar  baz
─────────────
1    2    3
```
What should `select(t, (:foo, Not(:bar))` do?

the options are:

1)
```julia
julia> select(t, (:foo, (:foo, :baz)))
Table with 1 rows, 2 columns:
1  2
─────────────────────
1  (foo = 1, baz = 3)
```
2)
```julia
julia> select(t, (:foo, :foo, :baz))
ERROR: syntax: duplicate field name: "foo" is not unique
```
I think it should be 1.

piever:
> A `(::OnAll)(x)`
> method would naturally make its use in `groupby` also work.

I see, this is probably a better approach. The name is maybe a bit unfortunate though, perhaps `ApplyColwise` ?

There is a bit of a problem with this approach in that returning a NamedTuple not always expands to different columns. See for example:

```julia
Main> groupby(t -> map(mean, columns(table(t))), iris, :Species, select = Not(:Species))
Table with 3 rows, 2 columns:
Species       #25
────────────────────────────────────────────────────────────────────────────────────────────────
"setosa"      (SepalLength = 5.006, SepalWidth = 3.418, PetalLength = 1.464, PetalWidth = 0.244)
"versicolor"  (SepalLength = 5.936, SepalWidth = 2.77, PetalLength = 4.26, PetalWidth = 1.326)
"virginica"   (SepalLength = 6.588, SepalWidth = 2.974, PetalLength = 5.552, PetalWidth = 2.026)
```

I'll try the callable type approach and see if I run into this problem (which is probably due to using `_promote_op`).

piever:
I've given this some more thoughts and looked how it's done in different environments (R's dplyr and Julia's DataFrames). They both have a specific function for this, called `aggregate` in Julia and `summarise_each` in dplyr that can accept grouped data. I believe we should also have our own function (`summarize` can be a good name, I don't like the `each` and the British spelling of dplyr). I think our general strategy of having `by` and `select` works beautifully here.

The syntax would then be:

```julia
summarize((mean, std), select(iris, Not(:Species))) # non grouped
summarize((mean, std), iris, :Species)
summarize(@NT(m = mean, s = std), iris, :Species) # If the user wants to use a different name for the new columns, useful when using anonymous functions
```

It has three advantages in my opinion:

- Closer to what other environments do / most familiar to users
- No problems with NamedTuple inference
- It can behave like `join` and by default only select columns that are not in `by` (i. .e `select = excludecols(t, by)`)

I can open a new PR with this implementation and we can see which one is better.

Concerning the special `select` syntax, I've realized that is less trivial than I thought, I'll take it away from this PR and will open an issue with my thoughts as well as your feedback.

piever:
Superseeded by #111

PR #111, opened 2018-01-05
Summarize all columns (second try)

piever:
I've tried a different implementation of #109 where I add a separate function `summarize`, that corresponds to dplyr's `summarise_each` and DataFrames' `aggregate`. It uses the usual `by`, `select` mechanism: see docstrings and tests for example usage.

### DIFF
--- a/src/reduce.jl
+++ b/src/reduce.jl
@@ -396,18 +396,6 @@ function groupby(f, t::Dataset, by=pkeynames(t); select=valuenames(t), flatten=f
         IndexedTables.flatten(t, length(columns(t))) : t
 end
 
-"""
-A callable type to apply functions column-wise to a table. Returns a `NamedTuple`.
-
-# Examples
-
-```jldoctest colwise
-julia> t = table([1, 2, 3], [1, 1, 1], names = [:x, :y]);
-
-julia> ApplyColwise(mean, std)(t)
-(x_mean = 2.0, y_mean = 1.0, x_std = 1.0, y_std = 0.0)
-```
-"""
 struct ApplyColwise{N, T<:Tuple}
     functions::T
     names::NTuple{N, Symbol}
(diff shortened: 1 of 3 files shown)


shashi:
I like this version a bit more.. will need a distributed version in JuliaDB, the code would look exactly like groupby

piever:
Ok, I'll close the other PR then, I also prefer it this way. The implementation uses `groupby` internally so it probably works already on distributed tables, I'll make a PR with docs and tests on JuliaDB.

piever:
The case with `by != ()` (where you split by some columns) works very easily in the distributed case, but I'm getting stuck in the `by == ()` case. What is the equivalent of `rechunk(t, by, select)` for `by == ()`?  That is to say, `groupby` seems to use `rechunk` to get to the `has_overlaps(t.domains, closed=true) == false` case. How do I do that when I do not have any key that I'm sorting by?

shashi:
Thank you!!

I've made a small tweak:

```julia
julia> x  = ndsparse(([1,1,1,1], [1,2,1,2]), @NT(x=[1,2,3,4], y=[5,6,7,8]), chunks=2)

julia> summarize(mean, x) # don't modify column name
2-d Distributed NDSparse with 2 values (2 field named tuples) in 1 chunks:
1  2 │ x    y
─────┼─────────
1  1 │ 2.0  6.0
1  2 │ 3.0  7.0

julia> summarize((mean,), x) # if you reaally want the _mean
2-d Distributed NDSparse with 2 values (2 field named tuples) in 1 chunks:
1  2 │ x_mean  y_mean
─────┼───────────────
1  1 │ 2.0     6.0
1  2 │ 3.0     7.0
```

shashi:
> How do I do that when I do not have any key that I'm sorting by?

This basically means we are to bring all the data onto a single process to finally apply `f`.

https://github.com/JuliaComputing/JuliaDB.jl/commit/6a0e6af263974178961e1304f10dae89e89fe38a

I also made the tweak to make summarize(f, ...) === groupby(ApplyColwise(f), ...) https://github.com/JuliaComputing/IndexedTables.jl/commit/53ca2e9710cc2a1b345212b8efa86335780c25d2

which allows me to just reexport `summarize` in JuliaDB.jl to provide the distributed version...

piever:
Nice! Be careful though, I think it's missing a `init_func(f::Function, data) = f` method for the normal `groupby` to work. I'll be adding an extra `init_func` dispatch on my JuliaDB PR to treat `DArray{<: Tup}` like `Columns` and not like `AbstractArray`

shashi:
You don't need that method though. Since JuliaDB only calls the serial `groupby` on chunks...

piever:
Ups, good point. I've updated [#118](https://github.com/JuliaComputing/JuliaDB.jl/pull/118) to only have docs and tests (plus a fix for `groupby` symbols or integers, as calling `isempty` on them errors).

PR #113, opened 2018-01-10
add missing init_func method

piever:
`groupby` is calling `f = init_func(f, data)` now, but the fallback for `init_func` was missing (see discussion in #111 ).


shashi:
Whoops! Thanks!!

PR #115, opened 2018-01-16
Widen a few methods: `Series` -> `AbstractSeries`

joshday:


shashi:
Does this require a version bump in REQUIRE?

joshday:
Yes, good catch!

PR #117, opened 2018-01-25
Fix ambiguous Series method

joshday:
Ref: https://github.com/JuliaLang/METADATA.jl/pull/13018#issuecomment-360324963


joshday:
Actually, @shashi how would you feel about simply removing this Series constructor?  I think I'd rather point people towards using `reduce`/`groupreduce`, which is a superset of functionality.

shashi:
That sounds good to me.

joshday:
Done.

PR #120, opened 2018-02-14
groupby revamped: column views and using key values

piever:
This packages implements already a `view` on `Columns`: it simply makes a new `Columns` object whose columns are views of the parent columns. I think this is what should be passed to `groupby`. Now it is much easier to select columns from this object.

For example:

```julia
julia> t = table(@NT(a = [1, 1, 2, 2], b = rand(4), c = rand(4)))
Table with 4 rows, 3 columns:
a  b           c
───────────────────────
1  0.00334594  0.195242
1  0.390544    0.612741
2  0.923973    0.243861
2  0.630879    0.795554

julia> groupby(t, :a) do dd
           mean(column(dd, :b).-column(dd, :c))
       end
Table with 2 rows, 2 columns:
a  #3
────────────
1  -0.207046
2  0.257718
```

In a later issue we should discuss whether we can also allow the user to use values from `key` rather than `data` [here](https://github.com/JuliaComputing/IndexedTables.jl/blob/master/src/reduce.jl#L259). For example Query allows you to do that with a special syntax `g.key`.

I didn't really understand the `SubArrayClosure` hack or whether it is required, but if it is I can simply do a `ViewClosure` type and recreate it.

### DIFF
--- a/src/reduce.jl
+++ b/src/reduce.jl
@@ -250,12 +250,6 @@ end
 
 ## GroupBy
 
-struct SubArrClosure{R}
-    r::R
-end
-
-(f::SubArrClosure)(x) = SubArray(x, f.r)
-
 function _groupby(f, key, data, perm, dest_key=similar(key,0),
                   dest_data=nothing, i1=1)
     n = length(key)
@@ -265,12 +259,10 @@ function _groupby(f, key, data, perm, dest_key=similar(key,0),
         while i <= n && roweq(key, perm[i], perm[i1])
             i += 1
         end
-        # needed this hack to avoid allocations. i loses type info
-        #val = _apply(f, map(x->SubArray(x, (perm[i1:(i-1)],)), cs))
         if isa(cs, Tup)
-            val = _apply(f, map(SubArrClosure((perm[i1:(i-1)],)), cs))
+            val = _apply(f, map(t -> view(t, perm[i1:(i-1)]), cs))
         else
-            val = _apply(f, SubArray(cs, (perm[i1:(i-1)],)))
+            val = _apply(f, view(cs, perm[i1:(i-1)]))
         end
 
         push!(dest_key, key[perm[i1]])
@@ -583,4 +575,3 @@ function reducedim_vec(f, x::NDSparse, dims; with=valuenames(x))
 end
 
 reducedim_vec(f, x::NDSparse, dims::Symbol) = reducedim_vec(f, x, [dims])
-
--- a/src/reduce.jl
+++ b/src/reduce.jl
@@ -250,20 +250,25 @@ end
 
 ## GroupBy
 
+_apply_with_key(f::Tup, data::Tup, process_data) = _apply(f, map(process_data, data))
+_apply_with_key(f::Tup, data, process_data) = _apply_with_key(f::Tup, columns(data), process_data)
+_apply_with_key(f, data, process_data) = _apply(f, process_data(data))
+
+_apply_with_key(f::Tup, key, data, process_data) = _apply_with_key(f, key, columns(data), process_data)
+_apply_with_key(f::Tup, key, data::Tup, process_data) = _apply(f, map(t->key, data), map(process_data, data))
+_apply_with_key(f, key, data, process_data) = _apply(f, key, process_cs(data))
+
 function _groupby(f, key, data, perm, dest_key=similar(key,0),
-                  dest_data=nothing, i1=1)
+                  dest_data=nothing, i1=1; usekey = false)
     n = length(key)
-    cs = f isa Tup ? columns(data) : data
     while i1 <= n
         i = i1+1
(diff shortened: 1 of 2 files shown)


davidanthoff:
I plan to change the story around ``.key`` in Query: I think folks will have to use a function ``key`` on the group to get the same value. The logic is this: I want to overload ``getproperty`` for the ``Grouping`` type once 0.7 lands to return columns. I think the equivalent here would be that the following type of syntax would work:
````julia
groupby(t, :a) do dd
    mean(dd.b .- dd.c) + key(dd)
end
````
``dd.b`` and ``dd.c`` would dispatch to ``getproperty`` and return the ``b`` and ``c`` column respectively. ``key(dd)`` would return the key for the current group.

**Comment by piever** [comment]
Thanks a lot for the feedback @davidanthoff. There is a practical concern though: in the case where for example only a column is selected, `dd` would be a normal array (which does not have any `key` information), so I'm not sure how something like `key(dd)` could work. I don't think we want to change that as it's actually very handy.

Is there any Julia magic that let's you use some special type/function that it will replace with the correct thing? I couldn't think of anything sensible...

An easier to implement alternative is something like:

```julia
groupby(t, :a, usekey = true) do (key, dd)
    mean(column(dd, :b).-column(dd, :c)) + key
end
```

where basically `usekey` determines whether you'll need to use information from `key` or not. If you need to use it, then you should use an anonymous function whose first argument is the `key` and the second the `data`. Once we do this, we may also get rid of having also the `key` columns be included in `data` by default.

davidanthoff:
In Query you always get a ``Grouping`` instance per group. ``Grouping`` inherits from ``AbstractArray``, so it is essentially an array that also has a key. So if you select only one column, you get an array with that element. If you select multiple columns, you get an array of named tuples. The ``dd.a`` key is essentially a shortcut syntax for ``map(i->i.a, dd)``.

There is one more thing I plan to do: there will be an ``AbstractGrouping`` type, and then the current ``Grouping`` type, plus a new one that stores named tuples as a struct of vectors, i.e. column wise. The ``getproperty`` implementation for the latter one will just return the right vector, i.e. that will be a lot faster.

shashi:
> SubArrayClosure hack

it could have been done with a vanilla closure but that was causing many allocations per row. Using this trick fixed that.

I feel `usekey` is a good way to pass in the key. I like its ergonomics the best!

piever:
>> SubArrayClosure hack

> it could have been done with a vanilla closure but that was causing many allocations per row. Using this trick fixed that.

I see, then it is maybe unnecessary with the new approach as it is not row based (I'm simply taking a view of each column vector).

> I feel `usekey` is a good way to pass in the key. I like its ergonomics the best! And it's also pretty non-intrusive.

Agreed. The only thing that would be breaking is that if the `key` values can be accessed with `usekey`, then the default column selection in `groupby` should be `select = t isa NDSparse ? valuenames(t) : Not(by)`.

If we agree on this `usekey` solution I can probably add it in this PR as it should be a reasonable small diff.

piever:
It was actually a slightly more extended diff than I originally thought.

Here is the state of this PR:

- `f` in `groupby` is fed a `Columns` `view` if multiple columns are selected (meaning a `Columns` of `SubArrays` of the original `Columns` object)
- the option `usekey=false` is provided. If set to true the user has to provide a function `f` that takes two arguments: `f(key, data)`
- the default column selection for NextTable is `Not(by)`, meaning all columns that are not indexing. Indexing columns do not need to be selected as they should be accessed with `usekey=true`

Should be in a state to be reviewed, let me know if you're happy with the changes.

davidanthoff:
This is probably irrelevant for the discussion here, but just for context, here is the reason why the key is part of the ``Grouping`` type in Query.jl: the groupby operation here in this package is essentially a combined groupby and projection operation as understood in Query. Now, to keep things composable in Query, we need to make sure that the groupby operation returns an iterable monad, i.e. it must return an iterator, i.e. a sequence of values. I could have returned a tuple of ``(key, groups)`` of course, but these are kind of unwieldy to handle because one would have to deconstruct these inside say a projection function manually, and if this is then composed into a longer sequence of query operators it all kind of breaks down.

piever:
I've actually used `groupby` as implemented in this PR quite a bit in the past couple of days and it works quite well: iterating through the `view` of `Columns` is the same as the way it was before, but selection of columns became much easier. Will merge in a couple of days if there are no concerns.

shashi:
This is a great change. I think one concern is the change here to use `Not(by)` as default `select` is breaking. We should create a NEWS.md and add this or something.

What happens if `Not(by)` is a single column?

shashi:
Okay the behavior here seems good to me. I'm merging this and will release with the NEWS.

piever:
Thanks! `Not(by)` always returns a `Tuple` when selected, even if only one column was excluded from `by`.

PR #127, opened 2018-02-26
allow column selection by regexes

piever:
See comment [here](https://github.com/JuliaComputing/JuliaDB.jl/pull/141#discussion_r170660525).

### DIFF
--- a/src/columns.jl
+++ b/src/columns.jl
@@ -571,7 +571,7 @@ struct Between{T1 <: Union{Int, Symbol}, T2 <: Union{Int, Symbol}}
     last::T2
 end
 
-const SpecialSelector = Union{Not, All, Keys, Between, Function}
+const SpecialSelector = Union{Not, All, Keys, Between, Function, Regex}
 
 lowerselection(t, s)                     = s
 lowerselection(t, s::Union{Int, Symbol}) = colindex(t, s)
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
You’ll want to be careful since regexes match any substring of a name which may not be the most intuitive default in this case.

piever:
My concern (in selecting only columns where the whole name is matched)  would be the inconsistence with `filter` which accepts matching just a substring:

```julia
julia> filter(r"a", ["ab"])
1-element Array{String,1}:
 "ab"
```

[dplyr](https://rdrr.io/cran/dplyr/man/select_helpers.html) also accepts any match (not necessarily matching the full name).
Maybe we can leave as is but document very clearly the behavior and explicitly recommend to use `^` and `$` if one wants to match the whole name.

Slightly off-topic, It was actually interesting to go through dplyr select helpers documentation, to have a reference point of how things are implemented there / what we're missing. We chose to allow any function to filter on names, but it seems dplyr only allows a few that are special cased:

- `starts_with()`: starts with a prefix (`x -> startswith(string(x), s)` for us or `r"^s"`)

- `ends_with()`: ends with a prefix (`x -> endswith(string(x), s)` for us or `r"s$"`)

- `contains()`: contains a literal string (`x -> contains(string(x), s)` for us or `r"s"`)

- `matches()`: matches a regular expression (just `r"s"` for us)

- `num_range()`: a numerical range like `x01, x02, x03`. (I guess simply `r"x\d+"` or `Tuple(Symbol.(x, range))`)

- `one_of()`: variables in character vector (we can pass the `Tuple` or use `x -> x in cols` if we want to allow the vector of names to contain names that are not columns of the dataset)

- `everything()`: all variables (`All()` for us)

It looks like we're doing a good job. Would it make sense to add `startswith`, `endswith` and `contains` methods also for `Symbols` to simplify the first few examples, or that's just too much for this small use case? Once there's curryfication + methods for symbols, it'd be `startswith(_, s)` which is much nicer.

shashi:
Cheers! :+1:

PR #128, opened 2018-02-27
allow setindex! on arrays with shared data

shashi:
This is a hack to detect if a buffer is shared. If it is it copies the NDSparse and then does flush!.


davidanthoff:
Is this effectively introducing copy-on-write semantics?

shashi:
No, not always. Only if your table was, say, memory-mapped.

davidanthoff:
Ah, ok, then I just misunderstood. Not sure about the second question :)

PR #133, opened 2018-03-17
flatten usability improvements

piever:
A couple of usability improvements for `flatten`:

- Default to last column. It allows `groupby(identity, t, col) |> flatten`
- Do not error if eltype of column to flatten is non-iterable, but simply return the same table. This allow `groupby(f, t, col, flatten = true)` to work with greater generality.

### DIFF
--- a/src/flatten.jl
+++ b/src/flatten.jl
@@ -164,9 +164,10 @@ function _flatten!(others, vecvec, out_others, out_vecvec)
 end
 
 """
-`flatten(t::Table, col)`
+`flatten(t::Table, col=length(columns(t)))`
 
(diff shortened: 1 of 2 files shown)


src/flatten.jl
```diff
@@ -201,8 +202,9 @@ x  a  b
 ```
 
 """
-function flatten(t::NextTable, col; pkey=nothing)
+function flatten(t::NextTable, col=length(columns(t)); pkey=nothing)
     vecvec = rows(t, col)
+    method_exists(start, (eltype(vecvec),)) || return t
```
shashi: pretty cool way of fixing this. Cheers!

PR #134, opened 2018-03-17
indexing usability improvements

piever:
This allows to call `getindex` and `view` with logical indexing (i.e. `t[[true, false]]`), whereas before it errored due to indexes being unsorted (which is incorrect, as this kind of indexing is always sorted).

I also allow unsorted indices if there are no primary keys. If there are primary keys, I can't quite decide whether we should remove those keys or error, I've left the error so far.


shashi:
I think erroring is fine for now. This is nice, thanks!

Issue #101, opened 2017-11-30
Tuple comprehension does not work with map


piever:
I found a mysterious bug: in `map`, if I use a function outputting a tuple that I write explicitly, everything works. If instead the tuple is a result of a tuple comprehension, it only stores the first column.

```julia
julia> t = table(rand(10), rand(10), names = [:x, :y])
Table with 10 rows, 2 columns:
x         y
───────────────────
0.6323    0.0750121
0.315025  0.884202
0.419645  0.55112
0.493855  0.822601
0.276584  0.71265
0.377394  0.399386
0.134123  0.574975
0.623424  0.399668
0.69115   0.0666205
0.310245  0.254196

julia> map(i -> (i.x, i.y), t)
Table with 10 rows, 2 columns:
1         2
───────────────────
0.6323    0.0750121
0.315025  0.884202
0.419645  0.55112
0.493855  0.822601
0.276584  0.71265
0.377394  0.399386
0.134123  0.574975
0.623424  0.399668
0.69115   0.0666205
0.310245  0.254196

julia> s = (:x, :y)
(:x, :y)

julia> map(i -> Tuple(getfield(i, j) for j in s), t)
Table with 10 rows, 1 columns:
1
────────
0.6323
0.315025
0.419645
0.493855
0.276584
0.377394
0.134123
0.623424
0.69115
0.310245
````

**Comment by shashi** [comment]
```julia
julia> IndexedTables._promote_op(i -> Tuple(getfield(i, j) for j in s), typeof(t[1]))
Tuple{Any}

```
I think in this case we can't infer the output type of the lambda. The right thing to do here is return an `Array{Any}`.

Use `astuple` to convert a named tuple to a tuple (it's a generated function, hence can be inferred)

```julia
julia> map(IndexedTables.astuple, t)
Table with 10 rows, 2 columns:
1          2
────────────────────
0.0824641  0.69732
0.953091   0.722422
0.335519   0.0399266
0.368203   0.319559
0.854868   0.783442
0.208625   0.961055
0.498797   0.347901
0.948223   0.433926
0.370459   0.808366
0.297149   0.0867216

```

piever:
Thanks for looking into this and for the tip! My concern came from the following inconsistence:

```julia
julia> map(i -> Tuple(getfield(i, j) for j in s), t)
10-element Array{Any,1}:
 (0.205925, 0.290762) 
 (0.720414, 0.145541) 
 (0.992084, 0.709543) 
[...]

PR #135, opened 2018-03-18
WIP: collect without inference

piever:
Collect an iterable of `Tuples` or `NamedTuples` to a `Columns` object, without using inference. The `Columns` object is initialized with the type of the first element and progressively widened, should this type change.
I assume that the iterator iterates `Tuples` (or `NamedTuples`) always of the same length and, in the `NamedTuples` case, always with the same fieldnames.

The condition to trigger widening is that at least one field of the iterate tuple is not a subtype of the `eltype` of the corresponding  column.

Widening happens element-wise, meaning I perform `typejoin` on a field by field basis.

To work with `missing` in the future `typejoin` is not the correct operation, as:

```julia
julia> typejoin(Missing, Int)
Any
```

but `Base.promote_typejoin` should be sufficient to fix this.

The overall strategy is more or less a copy-paste of Base Julia strategy to collect generators of unknown type.

### DIFF
--- a/src/collect.jl
+++ b/src/collect.jl
@@ -26,6 +26,32 @@ function collect_to_columns!(dest::AbstractArray{T}, itr, offs, st) where {T}
     return dest
 end
 
+function collectcolumns(itr, ::Base.SizeUnknown)
+    st = start(itr)
+    el, st = next(itr, st)
+    dest = similar(arrayof(typeof(el)), 1)
+    dest[1] = el
+    grow_to_columns!(dest, itr, 2, st)
+end
+
+function grow_to_columns!(dest::AbstractArray{T}, itr, offs, st) where {T}
+    # collect to dest array, checking the type of each result. if a result does not
+    # match, widen the result type and re-dispatch.
+    i = offs
+    while !done(itr, st)
+        el, st = next(itr, st)
+        if fieldwise_isa(el, T)
+            push!(dest, el)
+            i += 1
+        else
+            new = widencolumns(dest, i, el, T)
+            push!(new, el)
+            return grow_to_columns!(new, itr, i+1, st)
+        end
+    end
+    return dest
+end
+
 @generated function fieldwise_isa(el::S, ::Type{T}) where {S<:Tup, T}
     if all((s <: t) for (s, t) in zip(S.parameters, T.parameters))
         return :(true)
--- a/src/collect.jl
+++ b/src/collect.jl
@@ -4,7 +4,7 @@
 Collect an iterable as a `Columns` object if it iterates `Tuples` or `NamedTuples`, as a normal
 `Array` otherwise.
 
-## Example
+## Examples
 
 ```jldoctest collect
 julia> s = [(1,2), (3,4)];
--- a/src/collect.jl
+++ b/src/collect.jl
@@ -14,11 +14,11 @@ function collect_to_columns!(dest::Columns{T, U}, itr, offs, st) where {T, U}
     i = offs
     while !done(itr, st)
         el, st = next(itr, st)
-        S = typeof(el)
-        if all((s <: t) for (s, t) in zip(S.parameters, T.parameters))
+        if fieldwise_isa(el, T)
             @inbounds dest[i] = el::T
             i += 1
         else
+            S = typeof(el)
             Rparams = map(typejoin, T.parameters, S.parameters)
             R = get_tuple_type_from_params(el, Rparams)
             new = similar(arrayof(R), length(itr))
@@ -32,3 +32,11 @@ end
 
 get_tuple_type_from_params(el::Tuple, params) = Tuple{params...}
 get_tuple_type_from_params(el::NamedTuple, params) = eval(:(NamedTuples.@NT($(keys(el)...)))){params...}
+
+@generated function fieldwise_isa(el::S, ::Type{T}) where {S, T}
+    if all((s <: t) for (s, t) in zip(S.parameters, T.parameters))
+        return :(true)
+    else
+        return :(false)
+    end
+end
--- a/src/collect.jl
+++ b/src/collect.jl
@@ -1,0 +1,34 @@
+collectcolumns(itr) = collectcolumns(itr, Base.iteratorsize(itr))
+
+function collectcolumns(itr, ::Union{Base.HasShape, Base.HasLength})
+    st = start(itr)
+    el, st = next(itr, st)
+    dest = similar(arrayof(typeof(el)), length(itr))
+    dest[1] = el
+    collect_to_columns!(dest, itr, 2, st)
+end
+
+function collect_to_columns!(dest::Columns{T, U}, itr, offs, st) where {T, U}
+    # collect to dest array, checking the type of each result. if a result does not
+    # match, widen the result type and re-dispatch.
+    i = offs
+    while !done(itr, st)
+        el, st = next(itr, st)
+        S = typeof(el)
+        if all((s <: t) for (s, t) in zip(S.parameters, T.parameters))
+            @inbounds dest[i] = el::T
+            i += 1
+        else
+            Rparams = map(typejoin, T.parameters, S.parameters)
+            R = get_tuple_type_from_params(el, Rparams)
+            new = similar(arrayof(R), length(itr))
+            @inbounds for l in 1:i-1; new[l] = dest[l]; end
+            @inbounds new[i] = el
+            return collect_to_columns!(new, itr, i+1, st)
+        end
+    end
+    return dest
+end
+
+get_tuple_type_from_params(el::Tuple, params) = Tuple{params...}
+get_tuple_type_from_params(el::NamedTuple, params) = eval(:(NamedTuples.@NT($(keys(el)...)))){params...}
--- a/src/collect.jl
+++ b/src/collect.jl
@@ -1,6 +1,6 @@
-collectcolumns(itr) = collectcolumns(itr, Base.iteratorsize(itr))
+collect_columns(itr) = collect_columns(itr, Base.iteratorsize(itr))
 
-function collectcolumns(itr, ::Union{Base.HasShape, Base.HasLength})
+function collect_columns(itr, ::Union{Base.HasShape, Base.HasLength})
     st = start(itr)
     el, st = next(itr, st)
     dest = similar(arrayof(typeof(el)), length(itr))
@@ -26,7 +26,7 @@ function collect_to_columns!(dest::AbstractArray{T}, itr, offs, st) where {T}
     return dest
 end
 
-function collectcolumns(itr, ::Base.SizeUnknown)
+function collect_columns(itr, ::Base.SizeUnknown)
     st = start(itr)
     el, st = next(itr, st)
     dest = similar(arrayof(typeof(el)), 1)
--- a/src/collect.jl
+++ b/src/collect.jl
@@ -8,7 +8,7 @@ function collectcolumns(itr, ::Union{Base.HasShape, Base.HasLength})
     collect_to_columns!(dest, itr, 2, st)
 end
 
-function collect_to_columns!(dest::Columns{T, U}, itr, offs, st) where {T, U}
+function collect_to_columns!(dest::AbstractArray{T}, itr, offs, st) where {T}
     # collect to dest array, checking the type of each result. if a result does not
     # match, widen the result type and re-dispatch.
     i = offs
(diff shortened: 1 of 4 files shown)


src/collect.jl
```diff
@@ -0,0 +1,34 @@
...
+end
+
+function collect_to_columns!(dest::Columns{T, U}, itr, offs, st) where {T, U}
+    # collect to dest array, checking the type of each result. if a result does not
+    # match, widen the result type and re-dispatch.
+    i = offs
+    while !done(itr, st)
+        el, st = next(itr, st)
+        S = typeof(el)
+        if all((s <: t) for (s, t) in zip(S.parameters, T.parameters))
```
nalimilan: Is this inferable if `eltype(itr)` is concrete or once `T` has become at least as wide as it? That's essential for performance in Base `map`.

piever: I don't have a strong enough intuition to understand what the compiler can optimize. Here Base uses `el isa T || typeof(el) === T`. What should I check exactly to test inferrability? Maybe making this a function `is_fieldwise_subtype(el::S,  ::Type{T}) = all((s <: t) for (s, t) in zip(S.parameters, T.parameters))` would help the compiler, I'm not sure

piever:
Concerning missingness, I see the concern. One optimization I can do is to not copy the whole thing, but also figure out what is the "offending column" and only copy that one (there is a `setcol` function to only change one column without copying the rest: it should be pretty efficient). Like this each column is copied at most once. Of course, if there was a way to widen a vector to a vector that allows `missing`, this would be even better: is there any progress on that front?

src/collect.jl
nalimilan: `@code_warntype` will tell you whether the return type is inferred. Using a helper function could indeed help (maybe after marking it as `Base.@pure`, but the chances are that I'm wrong).

nalimilan:
> Concerning missingness, I see the concern. One optimization I can do is to not copy the whole thing, but also figure out what is the "offending column" and only copy that one (there is a setcol function to only change one column without copying the rest: it should be pretty efficient). Like this each column is copied at most once.

Good trick!

> Of course, if there was a way to widen a vector to a vector that allows missing without copying, this would be even better: is there progress on that front?

I don't think there's been any progress (the priority is merging breaking changes for 1.0).

> I'm starting to think that using the trick of "columnar storage" the situation is not nearly as bad as with a vector of NamedTuples, if copying each column at most once is considered acceptable, we may be able to skip inference altogether (plus, with your promote_typejoin PR, we wouldn't really need to special case the code so much to accomodate Missing).

Yeah, it should be more reasonable, at least if we are able to make `convert` non-copying. That's an interesting approach to test.

src/collect.jl
piever: I've checked with constant `eltype` and now it is tyoe stable. Somehow the way to achieve that was to use a generated function to check `is_fieldwise_subtype`, which we should do anyway for performance. It's not inferrable in the `NamedTuples` case (the initialization step `arrayof` which we already use in `map` is not inferrable), but I hope all the `NamedTuples` business will be much better in 0.7

piever: Still, should probably benchmark it to make sure it's doing OK compared with  the inference based `map`...

nalimilan: There's probably a way to achieve the same result without a generated function, but these things are always tricky...

piever:
Benchmarks are very encouraging, for some reason this is actually faster than our current implementation in the type stable case....

```julia
julia> using IndexedTables, BenchmarkTools

julia> s = table(rand(1000), rand(1000), names = [:x, :y]);

julia> f(i) = @NT(sum = i.x + i.y, diff = i.x - i.y)
f (generic function with 1 method)

julia> map1(f, s) = table(collectcolumns(f(i) for i in s), copy = false, presorted = true)
map1 (generic function with 1 method)

julia> @benchmark map($f, $s)
BenchmarkTools.Trial: 
  memory estimate:  105.02 KiB
  allocs estimate:  2554
  --------------
  minimum time:     64.973 μs (0.00% GC)
  median time:      72.453 μs (0.00% GC)
  mean time:        83.847 μs (7.97% GC)
  maximum time:     2.059 ms (91.21% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark map1($f, $s)
BenchmarkTools.Trial: 
  memory estimate:  17.83 KiB
  allocs estimate:  44
  --------------
  minimum time:     13.858 μs (0.00% GC)
  median time:      15.307 μs (0.00% GC)
  mean time:        17.561 μs (4.52% GC)
  maximum time:     1.198 ms (91.14% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> map1(f, s) == map(f,s)
true
```

I still need to:

- [ ] Implement the unknown length iterator case

- [ ] Generalize to work with scalars and just collect an `Array`

- [ ] Decide on a name for `collectcolumns` (though maybe `collectcolumns` is not so bad)

And then we could start porting things to use this instead of inference.

shashi:
That's pretty awesome! The patch looks good to me!! :-)

Now the challenge is figuring out how to implement `groupreduce` and `groupby` as iterators

Then there is the elephant in the room --- `join` :-)

src/collect.jl
```diff
@@ -0,0 +1,87 @@
...
+    end
+end
+
+@generated function fieldwise_isa(el::S, ::Type{T}) where {S, T}
+    if S <: T
+        return :(true)
+    else
+        return :(false)
+    end
+end
```
shashi: I think this doesn't need to be `@generated`, a `Base.@pure` or `@inline` might work here. But I guess once a package has `@generated` it really doesn't matter how many of them are there :-p

src/collect.jl
```diff
@@ -0,0 +1,87 @@
...
+function widencolumns(dest, i, el::S, ::Type{T}) where{S <: Tup, T}
+    sp, tp = S.parameters, T.parameters
+    idx = find(!(s <: t) for (s, t) in zip(sp, tp))
+    new = dest
+    for l in idx
+        newcol = Array{typejoin(sp[l], tp[l])}(length(dest))
+        copy!(newcol, 1, column(dest, l), 1, i-1)
+        new = setcol(new, l, newcol)
+    end
+    new
```
shashi: really liking how this function reads.

shashi:
> but `Base.promote_typejoin` should be sufficient to fix this.

I like the sound of that!

shashi:
Oops I think the current implementation of `map` allocates, probably because of https://github.com/JuliaLang/julia/issues/15276 since we close over `fs`?

shashi:
If I just call `f` instead of `fs` (which understandably isn't inferable), the performance is 20% slower in the relevant portions (allocation + filling), but I think we can take that for the benefits and simplicity this PR affords.

src/collect.jl
piever: `Base.@pure` seemed to not work as nicely, maybe we can leave as is now and do a "remove all generated functions" PR in the future (maybe in Julia 0.7).

piever:
I see, I couldn't really understand how this PR could be faster. 20% sounds fair: the last step of converting to a `table` could account for that. `map2(f,s) = collectcolumns(f(i) for i in s)` is much faster than `map1`, not sure why, I thought converting to `table` was for free.

In terms of `groupby` and `groupreduce`, I made the `SizeUnknown` implementation for them, after we merge this one I can see how to translate them to use it (`groupreduce` is already doing with this PR strategy but assuming concrete grouping). For `join` we should actually change `typejoin`.

We should (before 0.7, then there is `@nalimilan`'s PR to rescue us) define our own `promote_typejoin` and add `promote_typejoin(::Type{<:DataValue}, ::Type{T}) = DataValue{T}` and then it should work out of the box (though we may also need to tweak the `s <: t` part to allow `Float64 <: DataValue{Float64}` - of course using our own version of `<:`). Better leave it for last though.

A possible strategy could be, in the following order:

- Decide on a name for `collectcolumns` and merge this PR (could also leave `collectcolumns`)
- port `map`, which should be trivial
- port `groupby` and `groupreduce` which could be a bit trickier
- think about `join` and revisit this PR to adjust as needed.

The tricky thing about `group` is that it also uses inference to decide whether to wrap `f` in a function or not, so maybe that's not so easy.

shashi:
> I thought converting to table was for free.

That is suspicious actually, it's the simplest wrapper when `copy=false`, `presorted=true`... Worth profiling.

shashi:
I was not timing the table creation, and that could have been the 20% overhead... I did some profiling, it looks like it's just the overhead of dispatch calling many methods with keyword arguments.

shashi:
>  define our own `promote_typejoin` and add `promote_typejoin(::Type{<:DataValue}, ::Type{T}) = DataValue{T}`

```julia
julia> promote_type(DataValue{Int}, Float64)
DataValues.DataValue{Float64}
```

why not use `promote_type`?

I think the 4-step plan sounds perfect. `collectcolumns` seems like a good name. Maybe `collect_columns` is easier to read, and mirrors the use of _ after collect in Base.

Cheers!

piever:
`promote_type` is actually an interesting idea, I think I'll switch to that. I like it because it will give a tighter type than `typejoin`, then I guess we'll have to see in practice if that's enough to get `join` to work out of the box. I'll switch to that (and to `collect_columns` - I also think it reads better) + docstrings and I think it's good to go.

nalimilan:
One potential issue with `promote_type` is that `promote_type(Int, Float64) === Float64`, which means that big integers would be silently rounded. OTOH that's what happens with `cat` on arrays already, so...

piever:
I saw that but I think it's better than returning an `Array{Real}`, especially if there are precedents like `vcat`. The use case of mixing up large `Int64` and `Float64` while wanting to preserve the `Int64` exactly (where they are already at a size that risks overflow) seems limited.

shashi:
There's a `_promote_type` in JuliaDB.jl which does `promote_type` on all the fields of a tuple (using `map_params`). Maybe we should move that here and reuse it there. It's a minor detail that will need an in-tandem release. Everything looks good here: feel free to hit merge!

piever:
One thing though: throughout the code, I'm assuming that the number of fields in the output `Tup` is constant throughout the data, but one of the tests has `map` of a function which does not respect this property. The old code would give an `Array` of `Tuples` (as it is a case where inference failed). Should we just error in that case?

shashi:
Why can't we switch to `Array{Tuple}` in this case?

src/collect.jl
```diff
@@ -0,0 +1,113 @@
...
+
+@generated function fieldwise_isa(el::S, ::Type{T}) where {S, T}
+    if S <: T
+        return :(true)
+    else
+        return :(false)
+    end
+end
+
+function widencolumns(dest, i, el::S, ::Type{T}) where{S <: Tup, T}
```
shashi: I guess this should have `T <: Tup` as well?

src/collect.jl
```diff
@@ -0,0 +1,113 @@
...
+    sp, tp = S.parameters, T.parameters
+    idx = find(!(s <: t) for (s, t) in zip(sp, tp))
```
shashi: Would be good to see if `sp` and `tp` have the same length, else return `Array{Tuple}(length(dest)); copy!(newcol, 1, dest, 1, i-1)`?
`

piever: Yep, figured out that was missing when changing `map` and luckily one test checked for that. More than same length, same fieldnames (we also shouldn't accept `NamedTuples` where things are called differently). I'll change that in the future `map` PR.

piever: I should also add a test for #101 in the `map` PR as that should work now.

piever: It's actually not so trivial though, as the `Array{NamedTuple}` can get horribly inefficient (in the future `Missing` world), as separate fields can go missing in different moments and every time we'd need to copy everything (as the trick I'm doing here of only changing the relevant columns doesn't work). Not sure what's the best course of action: unless there are clear usecases it may be better to error.

piever:
It's actually not so trivial though, as the `Array{NamedTuple} ` can get horribly inefficient (in the future `Missing` world), as separate fields can go missing in different moments and every time we'd need to copy everything (as the trick I'm doing here of only changing the relevant columns doesn't work). Not sure what's the best course of action: unless there are clear usecases it may be better to error.

piever:
If the user insists of having tuples of different lengths, he/she can always wrap them in a one-element `Tuple`.

shashi:
> and every time we'd need to copy everything 

Why? `Array{NamedTuple}` should be wide enough for any tuple, even those containing missing?

piever:
Ah, fair enough, as soon as different length are encountered it will just give a `NamedTuple`. without specializing further. I'll see if there's a way to include this without making the code too ugly...

OTOH, don't you think it'd be better to error and propose that the user has to be explicit if he/she wants this funny `Array` of tuples of different length? `collect_columns((ntuple(identity, i),) for i in 1:3)`  for example.

davidanthoff:
You should be able to get rid of copying entirely if you widen to ``DataValueArray`` instead of ``Array{DataValue}`` (or are you doing that already? From a quick glance at the code I got the impression that you don't). In that case you can just reuse the array with the values that you collected so far, stuff it into a new DataValueArray and the only significant new allocation in a widening operation would be the missing mask (which of course is not a problem because it would have been allocated at some point in any case).

On the flip side, probably not even needed, because with ``DataValue`` you probably will pick the right column type right away based on the first element in the iteration in any case.

piever:
Thanks for the tip! It will probably be needed for `join` as there one could get `DataValue` because the row is missing even when the field is of type `Int` for example. I'll probably have to revisit this for `join` and will include your trick!

shashi:
We do the trick for DataValueArrays in TextParse. At some point (maybe with Parsing.jl) we should use this type of iterator approach there as well. The promotion rules there are quite complex (e.g. string columns are started off as PooledArray, then the pool is widened, if there are a large number of uniques, we switch to Array{String})

PR #137, opened 2018-03-20
map no longer relies on inference

piever:
I've addressed the last issues from #135 (allowing the user to give a mix of `NamedTuples` with different fields that gets stored as `Vector{NamedTuples}`) and update `map` to use the new `collect_columns`.

### DIFF
--- a/src/collect.jl
+++ b/src/collect.jl
@@ -79,6 +79,7 @@ function grow_to_columns!(dest::AbstractArray{T}, itr, offs, st) where {T}
 end
 
 # extra methods if we have widened to Vector{Tuple} or Vector{NamedTuple}
+# better to not generate as this is the case where the user is sending heterogenoeus data
 fieldwise_isa(el::S, ::Type{Tuple}) where {S<:Tup} = S <: Tuple
 fieldwise_isa(el::S, ::Type{NamedTuple}) where {S<:Tup} = S <: NamedTuple
 
--- a/src/collect.jl
+++ b/src/collect.jl
@@ -99,7 +99,11 @@ end
 end
 
 function widencolumns(dest, i, el::S, ::Type{T}) where{S <: Tup, T<:Tup}
-    if fieldnames(S) == fieldnames(T)
+    if fieldnames(S) != fieldnames(T) || S == Tuple || S == NamedTuple
+        R = (S <: Tuple) && (T <: Tuple) ? Tuple :  (S <: NamedTuple) && (T <: NamedTuple) ? NamedTuple : Any
+        new = Array{R}(length(dest))
+        copy!(new, 1, dest, 1, i-1)
+    else
         sp, tp = S.parameters, T.parameters
         idx = find(!(s <: t) for (s, t) in zip(sp, tp))
         new = dest
@@ -108,10 +112,6 @@ function widencolumns(dest, i, el::S, ::Type{T}) where{S <: Tup, T<:Tup}
             copy!(newcol, 1, column(dest, l), 1, i-1)
             new = setcol(new, l, newcol)
         end
-    else
-        R = (S <: Tuple) && (T <: Tuple) ? Tuple :  (S <: NamedTuple) && (T <: NamedTuple) ? NamedTuple : Any
-        new = Array{R}(length(dest))
-        copy!(new, 1, dest, 1, i-1)
     end
     new
 end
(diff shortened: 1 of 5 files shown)


piever:
Note that with the new `collect_columns` all sorts of new lazy iterator techniques become possible. For example

```julia
IndexedTables._map(i ->@NT(sum =  i.x+i.y, diff = i.x-i.y), Iterators.filter(f, t))
```

(we should maybe rename `_map` to `map_rows` and export it).

or equivalently

```julia
collect_columns(@NT(sum =  i.x+i.y, diff = i.x-i.y) for i in t if f(i))
```

which is a nice `Columns` comprehension. We may even use `Columns[...]` as a short-hand for that, though I'm not 100% sure which method would need to be overloaded for that.

Pinging @davidanthoff  on this one as I think it was one of his ideas to start implementing this kind of syntax.

piever:
Even better, we could consider defining `table(iterable; kwargs...)` as `table(collect_columns(iterable); kwargs...)` but we should figure out the best way to do so without clashing with the IterableTables definitions.

shashi:
What happens when table is called with a tuple or banned tuple of vectors then? Tuples are iterable too

davidanthoff:
Hehe, this is starting to look more and more like [Query.jl](https://github.com/davidanthoff/Query.jl) ;) Which really is nothing other than chained lazy iterators for everything (at least for the one backend we have right now).

About the `table` function: if you look at the [current](https://github.com/JuliaComputing/IndexedTables.jl/blob/11ffe1e186bf4f78a00ed533fc301707c7c718a6/src/tabletraits.jl#L69) implementation, I think the call to ``TableTraitsUtils.create_columns_from_iterabletable`` is probably pretty similar to ``collect_columns``? With the key difference of course that my implementation requires ``eltype``, whereas yours doesn't. I'll have to change my implementation eventually to mimic yours when I will purge the dependency on inference in [Query.jl](https://github.com/davidanthoff/Query.jl) (and if you don't mind I will probably just copy your code from here).

I think you should be able right now to just replace the whole function body with the following and it probably will just work with the whole iterable tables machinery:
````julia
function table(iter; kwargs...)
    return table(collect_columns(IteratorInterfaceExtensions.getiterator(iter)); kwargs...)
end
````
I haven't tried that, but I think the designs are essentially equivalent, so I'm optimistic it will just work. The motivation for the ``getiterator`` indirection is described [here](http://www.david-anthoff.com/jl4ds/latest/tabletraits.html#The-iterable-tables-interface-1).

Things might get a bit more complicated if IndexedTables moves over to ``Missing``. My best guess is still that julia 1.0 won't fix all the issues with ``Missing``, and if that is the case I will keep the iterable tables story on ``DataValue`` so that [Query.jl](https://github.com/davidanthoff/Query.jl) continues to work. In that case you would have to handle the case where a source has ``DataValue`` columns and do some conversions. Probably not really tricky, just something to keep in mind.

I'm also in the process of adding two more interfaces to [TableTraits.jl](https://github.com/davidanthoff/TableTraits.jl). Both of them are described in http://www.david-anthoff.com/jl4ds/latest/tabletraits.html. For now they should make certain scenarios much faster (initially probably mostly the file IO story, but we'll have to see). Eventually it might make sense to add support for those to the ``table`` function, but that should also be really easy and straightforward.

**Comment by piever** [comment]
@shashi the method to collect an iterable would be the one with the lowest dispatch preference, so it won't affect other methods. We may even consider whether we want a `Columns(iterable)` as well.

@davidanthoff the two advantages I see for the change are:

- no need for `eltype`
- we would accept also iterables of `Tuples` 

Nice job on the column based interface, we should definitely support it, it seems very useful also for IndexedTables to DataFrames conversion.

Feel free to copy paste this code (or change as needed) for IterableTables.

**Comment by piever** [comment]
I've added the `map_rows` function, which can take more than one argument after the function:

```julia
julia> map_rows(tuple, 1:3, ["a","b","c"])
3-element Columns{Tuple{Int64,String}}:
 (1, "a")
 (2, "b")
 (3, "c")
```

I think it may come in handy later for `join`.

Should be good to go on my side, I'll do the `table` iterator change in a separate PR.

piever:
@davidanthoff For the `table` construction from iterable, can I consider IteratorInterfaceExtensions stable and use `isiterable`?

shashi:
:rocket:

davidanthoff:
Oh, missed this question. Yes, IteratorInterfaceExtensions.jl is stable, I don't see any changes for that coming.

Issue #139, opened 2018-03-23
Proposal: `Pair` iterator to create tables with keys and to implement groupby/groupreduce


piever:
I've thought more about how to implement all the various IndexedTables operations in an "iterator" way, and it seems to me that `collect_columns` is not quite sufficient as most operations will be iterating both keys and values. Even for `map` it would be nice to have a way to denote that some columns of the output will be primary.

Here is my proposal:

- Special case `collect_columns` to the case where the `eltype` is `Pair{T1, T2} where {T1<:Tup, T2}`, which would be collected as a `Pair` of `Columns` (or a `Pair` of `Columns` and `Array` if `T2` is not a `Tup`. This is not too hard to do, one just needs to overload `Base.setindex!`, `Base.push!`, `Base.eltype` for `Pair{<:Columns, <:Any}` (which I hope is not type piracy as we own `Columns`)

- If necessary, add some methods `convert(::NextTable, ::Pair{<:Columns, <:Any}) ` and `convert(::NDSparse, ::Pair{<:Columns, <:Any}) `

- Reimplement `groupby`, `groupreduce` and `join` using iterators of `Pairs` of `Tup` and `collect_columns`

- In `map` accept functions that return `Pair` of `Tup`, in which case collect as table with primary keys

- In the `table(iterator)` constructor, accept iterators that iterate `Pair` of `Tup`, in which case collect as table with primary keys

davidanthoff:
Yeeeesss! I was just going to suggest this a couple of days ago (again, actually, see https://github.com/davidanthoff/IterableTables.jl/issues/26, which I believe is the same idea?).

davidanthoff:
Although, I think the iterators shouldn't return pairs, I think they should just continue to return named tuples, I think otherwise the syntax for most operations would get really weird. So I'm a huge fan of having ``collect`` style stuff accept pairs, but not to iterate pairs.

piever:
Yes, that's the same as your issue from last year :)

I also agree that one should iterate `NamedTuples`, not pairs...  In terms of implementation it may be useful to have a `ColumnsPair` object (though possibly not exported for now) that would iterate pairs and be used to collect this pairs, plus methods:

```julia
convert(NextTable, ColumnsPair)
convert(NDSparse, ColumnsPair)
```

that do not copy the data.

davidanthoff:
[...]

PR #140, opened 2018-03-26
Define and collect Columns{Pair}

piever:
Implementation of the proposal in #139. I would mainly use `Columns{Pair}` as an internal object without documenting it/recommending it to the user as  I have not yet implemented all the methods for `Columns` on `Columns{Pair}` (though most should be there already) and some I'm not sure whether it makes sense to implement at all. OTOH the most relevant methods are there and are enough for the `collect_columns` implementation to work for column pairs.

### DIFF
--- a/src/columns.jl
+++ b/src/columns.jl
@@ -325,6 +325,12 @@ function Base.vcat(c::Columns, cs::Columns...)
     Columns(map(vcat, map(x->x.columns, (c,cs...))...))
 end
 
+function Base.vcat(c::Columns{<:Pair}, cs::Columns{<:Pair}...)
+    Columns(vcat(c.columns.first, (x.columns.first for x in cs)...) =>
+            vcat(c.columns.second, (x.columns.second for x in cs)...))
+end
+
+
 abstract type SerializedColumns end
 
 function serialize(s::AbstractSerializer, c::Columns)
--- a/src/columns.jl
+++ b/src/columns.jl
@@ -188,7 +188,7 @@ julia> ncols(ndsparse(d, [7,8,9]))
 """
 function ncols end
 ncols(c::Columns) = nfields(typeof(c.columns))
-ncols(c::Columns{<:Pair, <:Pair}) = ncols(c.columns.first) + ncols(c.columns.second)
+ncols(c::Columns{<:Pair, <:Pair}) = ncols(c.columns.first) => ncols(c.columns.second)
 ncols(c::AbstractArray) = 1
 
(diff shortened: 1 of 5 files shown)


src/columns.jl
```diff
@@ -181,21 +188,23 @@ julia> ncols(ndsparse(d, [7,8,9]))
...
 
 function similar(c::Columns{D,C}) where {D,C}
-    cols = map(similar, c.columns)
+    cols = map_pair(similar, c.columns)
     Columns{D,typeof(cols)}(cols)
 end
 
 function similar(c::Columns{D,C}, n::Integer) where {D,C}
-    cols = map(a->similar(a,n), c.columns)
+    cols = map_pair(a->similar(a,n), c.columns)
```
shashi: maybe this should be called `_map` (this is the convention we use when we are avoiding piracy (e.g. _promote_op)). `map_pair` seems like it only works on pairs or something.

src/columns.jl
```diff
@@ -261,7 +275,10 @@ function sortperm(c::Columns)
     return p
 end
 
+sortperm(c::Columns{<:Pair}) = sortperm(c.columns.first)
```
shashi: Where is this used? Just checking if this is OK.

src/columns.jl
piever: Good point, I'll change that.

src/columns.jl
shashi: Hmm, I think we should probably remove this for now if it's not getting used anywhere. Maybe error messages will guide us to the right implementation.

piever: Well, but it allows `sort` and `sort!` to work out of the box with the `Columns{Pair}` (sorting by indexing `Columns`) and is in line with `issorted(c::Columns{<:Pair}) = issorted(c.columns.first)` which is quite useful to verify that our `keys => values` `Columns{Pair}` object is indeed sorted by `keys`.

shashi: If you sort a vector of pairs (`sort([1=>2,1=>1])`) it sorts by both keys and values. Hence my reluctance. We should use different names for these if possible.

piever: Good point, I'll change `sort` related functions to do that then. The sorting for pairs would then simply be lexicographic, right?

shashi: `sortperm(rows(c))` would do the trick if `rows(c)` strips the pairs... Or else you could use `refine_perm!` manually.

shashi:
LGTM!

PR #148, opened 2018-03-30
Construct table from pairs

piever:
This adds constructors and conversion methods to construct a `NextTable` or `NDSparse` from `Columns{<:Pair}` objects (the `first` field of the pair corresponding to primary keys), generalizes the iterator method for `table` to also accept iterators of tuples or pairs (in which case the `first` columns are primary, you can specify `presorted = true` if you know already they will be sorted).

`map` also allows functions that output a `Pair`, in which case the resulting table will have primary keys.

I think this should be the last tool needed to port `groupreduce` to the new `collect_columns` based implementation.

### DIFF
--- a/src/ndsparse.jl
+++ b/src/ndsparse.jl
@@ -209,6 +209,8 @@ function ndsparse(x::Columns, y::AbstractVector; kwargs...)
     ndsparse(columns(x), y; kwargs...)
 end
 
+ndsparse(c::Columns{<:Pair}; kwargs...) =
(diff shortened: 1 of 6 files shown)


piever:
Tests seem to be failing due to OnlineStats things I don't really understand, but they pass locally.

joshday:
Tests from OnlineStats will fail until its (pending green CI) latest metadata PR is merged

joshday:
All good now 👍

piever:
Perfect, thanks :)

shashi:
Cheers! :tada:

PR #149, opened 2018-03-30
Remove Series wrappers for OnlineStats

joshday:
Now that we have new releases of OnlineStats[Base], we no longer need to wrap stats in a Series.


shashi:
Does this require a version update in REQUIRE?

joshday:
Ah, yep.  Added it.

PR #150, opened 2018-04-02
move group algorithms to collect_columns technique

piever:
Also part of #139 . I've ported the `groupby` and `groupreduce` algorithm to the iterator approach. The core of the computation is now implemented in the `GroupBy` and `GroupReduce` iterators, but the algorithm being run is the same with two improvements:

- It no longer relies on inference
- If the returned object is not type stable, the column gets widened

The only caveat is that `groupreduce` with `f::Pair` used to work but I believe it was accidental (`f` was changed to `(f,)` if the inferred return type was not a `Tup`. I'm changing the test for that as I believe we only officially allow tuples or namedtuples of `Pair`. If you believe we should support `f::Pair` I can also do that.

I think the iterator approach allows some interesting future optimizations:

- `flatten` could be implemented directly on the iterator and we could say the intermediate step of collecting the data.
- we could add an option to return the `GroupBy` or `GroupReduce` iterator for "lazy grouping" if the user wants to combine it with other operations

### DIFF
--- a/test/test_core.jl
+++ b/test/test_core.jl
@@ -867,7 +867,7 @@ end
                     else
                         @NT y = x.y + y
                     end), t, :x, select=:z) == table([1, 2], [6, 15], names=Symbol[:x, :y])
-    @test groupreduce(:y => (+), t, :x, select=:z) == table([1, 2], [6, 15], names=Symbol[:x, :y])
+    @test groupreduce((:y => (+),), t, :x, select=:z) == table([1, 2], [6, 15], names=Symbol[:x, :y])
     t = table([1, 1, 1, 2, 2, 2], [1, 1, 2, 2, 1, 1], [1, 2, 3, 4, 5, 6], names=[:x, :y, :z])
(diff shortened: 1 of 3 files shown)


test/test_core.jl
```diff
@@ -867,7 +867,7 @@ end
                     else
                         @NT y = x.y + y
                     end), t, :x, select=:z) == table([1, 2], [6, 15], names=Symbol[:x, :y])
-    @test groupreduce(:y => (+), t, :x, select=:z) == table([1, 2], [6, 15], names=Symbol[:x, :y])
+    @test groupreduce((:y => (+),), t, :x, select=:z) == table([1, 2], [6, 15], names=Symbol[:x, :y])
```
shashi: I think it'd be nice to keep this, it's just a convenience thing. `groupby` should also support this...

piever: Ok, then I'll implement it properly for both, here it worked by chance (for example, it wouldn't have worked with a `NDSparse`).

piever: Fixed.

PR #153, opened 2018-04-04
Fixes collect_columns for pairs

piever:


src/collect.jl
```diff
@@ -87,19 +93,19 @@ end
 
 # extra methods if we have widened to Vector{Tuple} or Vector{NamedTuple}
 # better to not generate as this is the case where the user is sending heterogenoeus data
-fieldwise_isa(el::S, ::Type{Tuple}) where {S<:Tup} = S <: Tuple
-fieldwise_isa(el::S, ::Type{NamedTuple}) where {S<:Tup} = S <: NamedTuple
+fieldwise_isa(el::S, ::Type{Tuple}) where {S<:Tup} = _is_subtype(S, Tuple)
+fieldwise_isa(el::S, ::Type{NamedTuple}) where {S<:Tup} = _is_subtype(S, NamedTuple)
```
shashi: How is this different?

piever: In this case they are the same, but I thought it was better to be consistent and use a common `_is_subtype` function, so that as soon as we want to change something (for example when transitioning to `Missing`, we need to only change this one. The main use is `_is_subtype(Int, DataValue{Int})` is true, whereas `Int<:DataValue{Int}` gives false.

piever: I can of course change it back if you feel the previous version was cleaner. It's only important in the other two cases as the previous version led to a lot of unnecessary copying in some cases.

shashi: It's fine for now. In general we should add indirection only when required because following it should make some sense. :)

PR #154, opened 2018-04-04
More collect_column fixes

piever:
Correct function to choose whether to widen correctly in all cases, correctly create `DataValueArray` even if column starts as regular `Array` but widens later.

### DIFF
--- a/src/collect.jl
+++ b/src/collect.jl
@@ -4,6 +4,14 @@ _is_subtype(::Type{S}, ::Type{T}) where {S, T} = S <: T
 _is_subtype(::Type{S}, ::Type{DataValue{T}}) where {S, T} = S<:T
 _is_subtype(::Type{DataValue{S}}, ::Type{DataValue{T}}) where {S, T} = S<:T
 
+Base.@pure function dataarrayof(T)
+    if T<:DataValue
+        DataValueArray{T.parameters[1],1}
+    else
+        Vector{T}
+    end
+end
+
 """
 `collect_columns(itr)`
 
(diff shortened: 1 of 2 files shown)


src/collect.jl
```diff
@@ -1,8 +1,14 @@
...
-_is_subtype(::Type{DataValue{S}}, ::Type{DataValue{T}}) where {S, T} = S<:T
+_is_subtype(::Type{S}, ::Type{T}) where {S, T} = promote_type(S, T) == T
+
+Base.@pure function dataarrayof(T)
+    if T<:DataValue
+        DataValueArray{T.parameters[1],1}
+    else
+        Vector{T}
+    end
+end
```
shashi: There's already an `arrayof` function....

piever: But that one doesn't quite do it because it would create a `Columns` object when starting from a `Tuple`, which I don't think we want. For example `[((1,2), (2,3))]` should in my view give `Columns([(1,2)], [(2,3)])` rather than a nested columns object

shashi: Wouldn't Columns of Columns be more efficient storage? (e.g. in the case where the tuple contains strings.)

What makes it undesirable? It should have the same properties as a Vector of tuples.

piever: The more I think about it the more I agree with you, nested tuples should nest, `dataarrayof` can be replaced by `arrayof`, `fieldwise_isa` and `widencolumns` should be recursive and I'd need to add some tests for the nested corner cases.

PR #155, opened 2018-04-04
WIP/RFC: collect_columns_flattened

piever:
I'm adding a method to flatten and collect an iterator as `Columns`. It will be useful to optimize `groupby` with `flatten=true` (it will be done with just one pass over the data and the user will be able to save some allocations if the function in `groupby` returns an iterable rather than a table). I'm also needing it to convert `join` to the iterator approach.

It can either flatten an iterable of iterables or an iterable of pairs, where each pair is `key => iterable`.

Ideally it should have all the features of `flatten` and I hope could replace it eventually, but I'm not 100% sure what else `flatten` does other than flattening (in terms of naming or repeated keys).

I'm also not sure whether we want `collect_columns(x; flatten = false)`, it's handy but it seems we lose inferrability (though maybe that's fine). We could also want a "gentle" version that only flattens when it make sence (which could potentially be the one implemented as `collect_columns(x; flatten = false)`. The regular `flatten` checks that elements are iterable before attempting to flatten.

### DIFF
--- a/src/collect.jl
+++ b/src/collect.jl
@@ -36,14 +36,15 @@ julia> collect_columns(s)
   7
 ```
 """
-collect_columns(itr) = collect_columns(itr, Base.iteratorsize(itr))
+collect_columns(itr; flatten = false) =
+    collect_columns(itr, flatten ? Base.SizeUnknown() : Base.iteratorsize(itr); flatten = flatten)
 
 function collect_empty_columns(itr::T) where {T}
     S = Core.Inference.return_type(first, Tuple{T})
     similar(arrayof(S), 0)
 end
 
-function collect_columns(itr, ::Union{Base.HasShape, Base.HasLength})
+function collect_columns(itr, ::Union{Base.HasShape, Base.HasLength}; flatten = false)
     st = start(itr)
     done(itr, st) && return collect_empty_columns(itr)
     el, st = next(itr, st)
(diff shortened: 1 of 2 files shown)


shashi:
Do you think it would be possible to just use `Iterators.flatten` here?

It would seem like `Iterators.flatten` should compose with any iterator -- e.g. `GroupBy` to produce a stream of tuples...

piever:
That was my initial try as well, say `collect_columns(Iterators.flatten(x))` but somehow it seemed to me it was not very efficient and when implementing `join` I ended up with a very complex concatenation of iterators (especially because in the case `key => iterator` I need to do the further transformation `(key => i for i in iterator)` before flattening. I'll try to do some proper benchmarking to see if there really is some performance impact or I was imagining it.

piever:
It looks like this method is needed: there's a performance difference which becomes massive in the type unstable case (which we need for `join`). I think `flatten` is the reason why I wasn't managing to get decent performance with an iterator based `join`:

```julia
v = (zeros(10) for i in 1:100)
@btime collect_columns_flattened(v)
# 7.528 μs (109 allocations: 36.11 KiB)

@btime collect_columns(Iterators.flatten(v))
 # 14.669 μs (1115 allocations: 63.56 KiB)

v2 = (i > 50 ? zeros(10) : zeros(Int,10) for i in 1:100)
@btime collect_columns_flattened(v2)
#  8.599 μs (111 allocations: 38.02 KiB)

@btime collect_columns(Iterators.flatten(v2))
#  442.535 μs (2619 allocations: 106.58 KiB)
```

piever:
I'll definitely need to add tests before merging. And I think I'll need to understand better what our own `flatten` actually does (say in terms of column naming and duplicated keys) and do some more performance testing to see if we can use this one instead of `flatten` in `groupby` (so a user could return an iterable with `f` in `groupby` and we'd save a lot of allocations).

piever:
For example, I was thinking a sensible strategy for naming here would be to simply use the names of what is being iterated, except if I'm iterating scalars, in which case I could take the name of the original column.

PR #161, opened 2018-04-11
Column modification usability improvement

piever:
Still lacks docs and tests but I was wondering if it is something people find useful. The strategy is:

- For column modificators that take two arguments other than the table (name and value or name and new name) add a `Pair` option, i. e. `setcol(t, name => value)`
- Add iterator version `setcol(t, itr) = foldl(setcol, t, itr)` as well as `setcol(t, itr...)` with the same meaning
- Add `transformcol` that combines `setcol` and `pushcol`: modify a column if it exists, add it otherwise. I have it implemented in JuliaDBMeta but I thought it made sense to have it here.

Example usage:

```julia
julia> t = table(rand(10), rand(10), rand(10), names = [:a, :b, :c]);

julia> renamecol(t, :a => :aa, :b => :bb)
Table with 10 rows, 3 columns:
aa        bb         c
─────────────────────────────
0.163642  0.0324896  0.117211
0.926122  0.102359   0.560399
0.917097  0.590855   0.325438
0.215478  0.734164   0.77882
0.594439  0.702912   0.609979
0.812428  0.757228   0.627404
0.804359  0.621889   0.17073
0.510328  0.387981   0.673884
0.593012  0.591315   0.290553
0.892289  0.486617   0.142943

# Can also take iterable
julia> setcol(t, (:a => fill(1, 10), :b => fill(1, 10)))
Table with 10 rows, 3 columns:
a  b  c
──────────────
1  1  0.117211
1  1  0.560399
1  1  0.325438
1  1  0.77882
1  1  0.609979
1  1  0.627404
1  1  0.17073
1  1  0.673884
1  1  0.290553
1  1  0.142943

julia> transformcol(t, :a => fill(1, 10), :d => fill(1, 10))
Table with 10 rows, 4 columns:
a  b          c         d
─────────────────────────
1  0.0324896  0.117211  1
1  0.102359   0.560399  1
1  0.590855   0.325438  1
1  0.734164   0.77882   1
1  0.702912   0.609979  1
1  0.757228   0.627404  1
1  0.621889   0.17073   1
1  0.387981   0.673884  1
1  0.591315   0.290553  1
1  0.486617   0.142943  1
```

### DIFF
--- a/src/columns.jl
+++ b/src/columns.jl
@@ -920,8 +920,11 @@ function Base.push!(d::ColDict, key::Union{Symbol, Int}, x)
     push!(d.columns, rows(d.src, x))
 end
 
-for (s, typ) in zip([:(Base.pop!), :(Base.push!), :(rename!), :(set!)], [:(Union{Symbol, Int}), :Pair, :Pair, :Pair])
-    if s != :(Base.pop!)
+for s in [:(Base.pop!), :(Base.push!), :(rename!), :(set!)]
+    if s == :(Base.pop!)
+        typ = :(Union{Symbol, Int})
+    else
+        typ = :Pair
         @eval $s(t::ColDict, x::Pair) = $s(t, x.first, x.second)
     end
     @eval begin
@@ -1016,6 +1019,24 @@ julia> t == t2
 false
 
 ```
+
+If `col` is not an existing column, `setcol` will add it:
+
+```jldoctest setcol
+julia> t = table([1,2], [2,3], names = [:a,:b])
+Table with 2 rows, 2 columns:
+a  b
+────
+1  2
+2  3
+
+julia> setcol(t, :c, [1,2])
+Table with 2 rows, 3 columns:
+a  b  c
+───────
+1  2  1
+2  3  2
+```
 """
 setcol(t, args...) = @cols set!(t, args...)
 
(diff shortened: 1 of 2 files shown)


shashi:
+1 from me!

> Add `transformcol` that combines `setcol` and `pushcol`: modify a column if it exists, add it otherwise.

`setcol` already creates the column if it doesn't exist...

piever:
That's somewhat embarrassing... I'll remove `transformcol` and add docs for this feature of `setcol` then.

shashi:
while `foldl` is quite an elegant way of doing this, I'd recommend using the `ColDict` to the fullest -- `foldl` will create new types in each iteration and hence may be slow. `ColDict` will only create a new type when you call `dict[]`.

piever:
Ah, I see, because if I do `foldl` there I recreate the table, I should actually move the `foldl` part in `ColDict` space.

src/columns.jl
```diff
@@ -1015,6 +1018,24 @@ t     x  y
...
+────
+1  2
+2  3
+
+julia> setcol(t, :c, [1,2])
+Table with 2 rows, 3 columns:
+a  b  c
+───────
+1  2  1
+2  3  2
```
shashi: Is there a `setcol(t, cols::Pair...)`?

piever: Surprisingly it was documented already but not implemented :) So it does not show the documentation in the diff as it was there already and I just left it. But yes, `setcol(t, cols::Pair...)` works now, see [this test](https://github.com/JuliaComputing/IndexedTables.jl/pull/161/files#diff-1649e312fc01ff2d3b8d62d9a4e334c5R546).

shashi: LOL

PR #166, opened 2018-04-24
optimize `permute!` for certain data types

tanmaykm:
Optimize `permute!` of columns of types `StringVector{String}` and `PooledArray`

Without this change:

```
julia> using IndexedTables, WeakRefStrings, PooledArrays, BenchmarkTools

julia> sv = StringVector{String}(convert(Vector{UInt8}, randstring(10^6+10)), UInt64[1:10:10^6;], ones(UInt32,10^5)*9);

julia> pv = PooledArray([randstring(5) for idx in 1:10^5]);

julia> cs = Columns(@NT(s=sv));

julia> cp = Columns(@NT(a=pv));

julia> p = randperm(10^5);

julia> @btime permute!(cs, p);
  9.412 ms (200025 allocations: 9.50 MiB)

julia> @btime permute!(cp, p);
  31.896 ms (199530 allocations: 12.54 MiB)
```

With this change:

```
julia> @btime permute!(cs, p);
  1.553 ms (3 allocations: 781.36 KiB)

julia> @btime permute!(cp, p);
  1.454 ms (2 allocations: 781.33 KiB)
```

### DIFF
--- a/src/columns.jl
+++ b/src/columns.jl
@@ -314,10 +314,7 @@ end
 
 function permute!(c::Columns, p::AbstractVector)
     for v in c.columns
-        if isa(v, StringArray{String})
-            ws = convert(StringArray{WeakRefString{UInt8}}, v)
-            permute!(ws, p)
-        elseif isa(v, PooledArrays.PooledArray)
+        if isa(v, PooledArrays.PooledArray) || isa(v, StringArray{String})
             permute!(v, p)
         else
             copy!(v, v[p])
--- a/src/columns.jl
+++ b/src/columns.jl
@@ -314,7 +314,14 @@ end
 
 function permute!(c::Columns, p::AbstractVector)
     for v in c.columns
-        copy!(v, v[p])
+        if isa(v, StringArray{String})
+            ws = convert(StringArray{WeakRefString{UInt8}}, v)
+            permute!(ws, p)
+        elseif isa(v, PooledArrays.PooledArray)
+            permute!(v, p)
+        else
+            copy!(v, v[p])
+        end
     end
     return c
 end


src/columns.jl
```diff
@@ -314,7 +314,14 @@ end
...
 function permute!(c::Columns, p::AbstractVector)
     for v in c.columns
-        copy!(v, v[p])
+        if isa(v, StringArray{String})
+            ws = convert(StringArray{WeakRefString{UInt8}}, v)
+            permute!(ws, p)
+        elseif isa(v, PooledArrays.PooledArray)
+            permute!(v, p)
+        else
+            copy!(v, v[p])
```
shashi: I don't know why this was the implementation....

Shouldn't `permute!(v, p)` work in the general case?

src/columns.jl
```diff
@@ -314,7 +314,14 @@ end
 
 function permute!(c::Columns, p::AbstractVector)
     for v in c.columns
-        copy!(v, v[p])
+        if isa(v, StringArray{String})
+            ws = convert(StringArray{WeakRefString{UInt8}}, v)
+            permute!(ws, p)
```
shashi: Ideally this should be the implementation of this in WeakRefStrings.

tanmaykm: PR opened: https://github.com/JuliaData/WeakRefStrings.jl/pull/39.
Will remove from here once that's merged.

src/columns.jl
tanmaykm: I think because:
```
julia> using BenchmarkTools

julia> a = rand(Int, 10^6);

julia> p = shuffle!([1:10^6;]);

julia> @btime permute!(a, p);
  65.502 ms (2 allocations: 7.63 MiB)

julia> @btime copy!(a, a[p]);
  9.788 ms (3 allocations: 7.63 MiB)
```

shashi: makes sense it's probably because the cache keeps getting invalidated on every write if you don't use a separate buffer?

PR #167, opened 2018-04-26
keep merged columns as PooledArrays when possible

tanmaykm:
The default merge behavior of PooledArray is to always convert the result to a non pooled type.

This overrides that behavior when the non pooled column is of reasonable length.

### DIFF
--- a/src/join.jl
+++ b/src/join.jl
@@ -878,8 +878,15 @@ function Base.merge(a::NextTable, b::NextTable;
 end
 
 opt_vcat(a, b) = vcat(a, b)
-opt_vcat(a::AbstractArray{<:Any, 1}, b::PooledArray{<:Any, <:Integer, 1}) = vcat((length(unique(a)) <= length(b.pool)) ? PooledArray(a) : a, b)
-opt_vcat(a::PooledArray{<:Any, <:Integer, 1}, b::AbstractArray{<:Any, 1}) = vcat(a, (length(unique(b)) < length(a.pool)) ? PooledArray(b) : b)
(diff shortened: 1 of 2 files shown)


shashi:
Nice :+1:

src/join.jl
```diff
@@ -874,9 +874,13 @@ function Base.merge(a::NextTable, b::NextTable;
     else
         throw(ArgumentError("the tables don't have the same column names. Use `select` first."))
     end
-    table(map(vcat, columns(a), columns(b)), pkey=pkey, copy=false)
+    table(map(opt_vcat, columns(a), columns(b)), pkey=pkey, copy=false)
 end
 
+opt_vcat(a, b) = vcat(a, b)
+opt_vcat(a::AbstractArray{<:Any, 1}, b::PooledArray{<:Any, <:Integer, 1}) = vcat((length(unique(a)) <= length(b.pool)) ? PooledArray(a) : a, b)
+opt_vcat(a::PooledArray{<:Any, <:Integer, 1}, b::AbstractArray{<:Any, 1}) = vcat(a, (length(unique(b)) < length(a.pool)) ? PooledArray(b) : b)
```
shashi: Maybe it's worth writing an optimized version of

`length(unique(b)) < length(a.pool)`

which will stop trying to find uniques once `length(a.pool)` number of them have been found.

tanmaykm: yes, good idea.

tanmaykm: done in https://github.com/JuliaComputing/IndexedTables.jl/pull/167/commits/16e648f52586e5e7587c318ffb77f8d1d6d1c665

PR #170, opened 2018-05-03
keep WeakRefString in StringArray instead of Array

tanmaykm:
to ensure they are serialized properly

fixes https://github.com/JuliaComputing/JuliaDB.jl/issues/198 and https://github.com/JuliaComputing/JuliaDB.jl/issues/199


andreasnoack:
Great. Thanks for fixing these.

PR #172, opened 2018-05-09
optimize `sortperm` on Columns of StringArrays

shashi:
by converting them to `StringArray{WeakRefString{UInt8}}`

Are there any compile time implications of this? @JeffBezanson

### DIFF
--- a/src/columns.jl
+++ b/src/columns.jl
@@ -272,14 +272,14 @@ function sortperm(c::Columns)
     cols = c.columns
     x = cols[1]
     if (eltype(x) <: AbstractString && !(x isa PooledArray)) || length(cols) > 1
-        pa = PooledArray(x)
+        pa = PooledArray(compact_mem(x))
         p = sortperm_fast(pa)
     else
         p = sortperm_fast(x)
     end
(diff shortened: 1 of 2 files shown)


src/utils.jl
```diff
@@ -391,3 +391,7 @@ function isshared(x)
         end
     end
 end
+
+compact_mem(x) = x
+compact_mem(x::StringArray{String}) = convert(StringArray{WeakRefString{UInt8}}, x)
+compact_mem(x::Columns) = Columns(map(compact_mem, columns(x)))
```
shashi: Would `@noinline` be useful here?

src/columns.jl
```diff
@@ -1293,3 +1293,7 @@ function init_inputs(f::Tup, input, gettype, isvec)
     # functions, input, output_eltype
     NT(fs...), rows(NT(xs...)), NT{output_eltypes...}
 end
+
+### utils
+
+compact_mem(x::Columns) = Columns(map(compact_mem, columns(x)))
```
shashi: Would @noinline be useful here?

PR #180, opened 2018-06-24
Added ability to create dummy variables from categorical feature columns

DhairyaLGandhi:
Added support for creating dummy variables in tables where representing categorical features as such may be desired.

### DIFF
--- a/src/reshape.jl
+++ b/src/reshape.jl
@@ -27,30 +27,30 @@ x  variable  value
 ```
 """
 function stack(t::D, by = pkeynames(t); select = isa(t, NDSparse) ? valuenames(t) : excludecols(t, by),
-    variable = :variable, value = :value) where {D<:Dataset}
+	variable = :variable, value = :value) where {D<:Dataset}
 
-    (by != pkeynames(t)) && return stack(reindex(t, by, select); variable = :variable, value = :value)
+	(by != pkeynames(t)) && return stack(reindex(t, by, select); variable = :variable, value = :value)
 
-    valuecols = columns(t, select)
-    valuecol = [valuecol[i] for i in 1:length(t) for valuecol in valuecols]
+	valuecols = columns(t, select)
+	valuecol = [valuecol[i] for i in 1:length(t) for valuecol in valuecols]
 
-    labels = fieldnames(valuecols)
-    labelcol = [label for i in 1:length(t) for label in labels]
+	labels = fieldnames(valuecols)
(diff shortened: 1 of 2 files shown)


src/reshape.jl
```diff
@@ -95,3 +95,57 @@ function unstack(t::D, by = pkeynames(t); variable = :variable, value = :value)
...
+
+function create_dummies(t, cols; categories=Dict())
+    for i in cols
+        if haskey(categories, i)
+            uniq = categories[i]
+        else
+            uniq = collect(keys(reduce(CountMap(), t, select = i).value))
+        end
+        for j in uniq
+            t = setcol(t, Symbol("$(i)_$(j)"), i => x -> x == j ? Int8(1) : Int8(0))
```
piever: `setcol` can take a list of name => value pairs and it's more efficient as it doesn't create all the intermediate tables: `setcol(t, :a => col_a, :b => col_b)`

piever:
It looks good but I'm curious what's the use case: normally one would use StatsModels.jl to go from categorical variables to columns of `0`s and `1`s. Could whatever model needs this rely on StatsModels instead?

shashi:
I think StatsModels will give you a matrix from one column. This PR keeps the table structure. It's quite handy when you're building a table of features and eventually plan to pass `hcat(columns(t)...)` to fit, say, a decision tree. Otherwise you'd need to deal with each categorical feature (or the set of them) separately. I'm fine with having this here till we come up with a better solution that may or may not involve StatsModels.

src/reshape.jl
```diff
@@ -95,3 +95,58 @@ function unstack(t::D, by = pkeynames(t); variable = :variable, value = :value)
...
+        for j in uniq
+            push!(arr, Symbol("$(i)_$(j)") => i => f(j))
+        end
+    end
+    t = setcol(t, arr)
+    t = popcol(t, cols)
+    t
+end
+create_dummies(t, col::Symbol; categories = Dict()) = create_dummies(t, [col], categories = categories)
+f(j) = x -> x == j ? Int8(1) : Int8(0)
```
piever: Maybe a more expressive name, or define it within the `create_dummies` function: it's a bit strange to take up a one letter name globally in the package. As you use it only once and it's a very short function, it's probably best to simply use the anonymous function:
```julia
push!(arr, Symbol("$(i)_$(j)") => i => x -> Int8(x == j))
```

PR #182, opened 2018-07-11
port to Julia 0.7

shashi:
Replace DataValues.jl with `Union{Missing,T}` and NamedTuples.jl with native named tuples.

Get all but TableTraits.jl code to pass tests.

cc @davidanthoff I have commented tabletraits.jl code for now. Are you planning to make it possible to use it on 0.7 with `Union{Missing, T}`?

cc @quinnj 

@piever maybe `_is_subtype` is now kind of redundant since `Union{Missing, T}` can be tested with `<:` for the same property?


piever:
I think we should switch to `Base.promote_typejoin` to widen element when collecting an iterable of tuples, but @nalimilan probably knows more about this.

I would also like to get rid of `is_subtype`, but we need to make sure that `Base.promote_typejoin(S, T) === T` if and only if `S <: T`. If that's not the case we could even use `Base.promote_typejoin(S, T) !== T` as a condition to widen (meaning, widen only if it would actually get strictly wider).

nalimilan:
> I think we should switch to Base.promote_typejoin to widen element when collecting an iterable of tuples, but @nalimilan probably knows more about this.

Yes, that sounds right. Basically, do the same thing as `map`.

> I would also like to get rid of is_subtype, but we need to make sure that Base.promote_typejoin(S, T) === T if and only if S <: T. If that's not the case we could even use Base.promote_typejoin(S, T) !== T as a condition to widen (meaning, widen only if it would actually get strictly wider).

I would say that's the case. At least `promote_typejoin` is equivalent to `typejoin` except for `Union{T,Missing}` and `Union{T,Nothing}`.

src/table.jl
```diff
@@ -28,7 +27,7 @@ struct NextTable{C<:Columns} <: AbstractIndexedTable
     # Cache permutations by various subsets of columns
     perms::Vector{Perm}
     # store what percent of the data in each column is unique
-    cardinality::Vector{Nullable{Float64}}
+    cardinality::Vector{Any}
```
nalimilan: Should this be `Union{Float64,Missing}`?

.travis.yml
```diff
@@ -4,7 +4,7 @@ os:
   - linux
   - osx
 julia:
-  - 0.6
+  - nightly
```
andreasnoack: Should add `0.7` and `1.0` as well.

andreasnoack:
Bump. It would be great to get this merged soon.

JeffBezanson:
David's performance concern here is valid --- it might be better to do everything but the Missing change first.

shashi:
@davidanthoff I think @piever's iterator based approach with `widencolumns` was key to making this possible with `missing`. He did strategically design for that with this change in mind.

quinnj:
It would certainly be helpful if some benchmarks were actually performed. @piever, do you happen to have any comparisons you did in doing the `widencolumns` work?

shashi:
Of course. I was just commenting on the point about not being able to use Iterators + missing.

shashi:
Ah Okay I hadn't schooled myself on that discourse discussion, my bad.

davidanthoff:
@piever's code is great, but I wouldn't expect it to solve the performance problems. I might be wrong, so benchmarking and comparing would be a good idea.

piever:
At the time I mainly tested that it didn't lose performance compared to the previous inference based implementation in the type stable case: as `Missing` in Julia 0.6 was slow anyway I didn't test it. I suspect there may be issues in the case of multiple columns allowing missing data, with the combinatorial explosion of possible types, but I don't have benchmarks.

OTOH, I think that https://github.com/JuliaData/Tables.jl/pull/10 is managing to combine the good ideas from `collect_columns` and the good ideas from Tables and may be the ideal implementation: in case my implementation here has performance issues we could try using that one instead (I also think it's slightly suboptimal to have to maintain two separate implementations of the same thing). It would allow several other optimizations as well like iterating lazy rows (see [here](https://github.com/JuliaData/Tables.jl/blob/master/src/Tables.jl#L112)) rather than materializing `NamedTuples` like we currently do.

On a related note, I also seem to remember that some functions had yet to be ported to the new iteration framework (`join` for example) and still rely on inference, so we probably need to keep that in mind.

test/test_collect.jl
```diff
@@ -1,44 +1,44 @@
 using IndexedTables: collect_columns_flattened
 
 @testset "collectnamedtuples" begin
-    v = [@NT(a = 1, b = 2), @NT(a = 1, b = 3)]
-    @test collect_columns(v) == Columns(@NT(a = Int[1, 1], b = Int[2, 3]))
+    v = [(a = 1, b = 2), (a = 1, b = 3)]
+    @test collect_columns(v) == Columns((a = Int[1, 1], b = Int[2, 3]))
```
tshort: It'd be nice to have a constructor for `Columns` using keyword args, so a level of parens can be dropped.

```julia
@test collect_columns(v) == Columns(a = Int[1, 1], b = Int[2, 3])
```

JeffBezanson: I think there is such a constructor; maybe this is just checking that passing a (named)tuple also works.

davidanthoff:
>  I suspect there may be issues in the case of multiple columns allowing missing data, with the combinatorial explosion of possible types, but I don't have benchmarks.

Yes, that is the scenario I would worry about. Doesn't seem like a corner case to me :)

I think in general a good strategy would be to get this working on julia 0.7, with the minimal set of change, and then think about broader redesigns. No need to couple those two decisions and thereby delay everything by a lot.

quinnj:
Just pushed a commit that gets this further for 0.7/1.0, some remaining test failures involve:

* Some complaints about `isnull` not being defined: @davidanthoff is this a DataValues thing/change?
* Some complaints about no method for `setindex_shape_check(::Bool, ::Int64)`; I'm guessing that was an internal Base method that got changed?
* A bunch of join tests are still checking DataValue missing vs. `missing`
* `  MethodError: no method matching +(::Int64, ::UnitRange{Int64})` from `   [1] valuenames(::NDSparse{Tuple{Int64,Int64},Tuple{Int64,Int64},Columns{Tuple{Int64,Int64},Tuple{Array{Int64,1},Array{Int64,1}}},Columns{Tuple{Int64,Int64},Tuple{Array{Int64,1},Array{Int64,1}}}}) at /Users/jacobquinn/.julia/dev/IndexedTables/src/ndsparse.jl:271`
* And a PooledArray error like `  MethodError: no method matching copy!(::Array{String,1}, ::Int64, ::PooledArray{String,UInt8,1,Array{UInt8,1}}, ::Int64, ::Int64)`.

Can anybody else pick things up from here? If not, I can try to take another crack or two over the next few days.

piever:
> Some complaints about isnull not being defined: @davidanthoff is this a DataValues thing/change?

This is also affecting StatPlots, FWIU it is `isna` now.

JeffBezanson:
The `+` should be `.+` and `copy!` should be `copyto!`.

davidanthoff:
Yes, ``isna`` is the new ``isnull`` on 0.7+ for ``DataValue``.

I haven’t gotten the broadcasting stuff to work for DataValues.jl on 0.7 yet. Not sure that is used anywhere here, though, just a heads up.

shashi:
BTW I did a simple map with tables of n missable/DataValue columns. It does scale exponentially with Missing...

Nice work @quinnj ! I'll try to fix up the tests tomorrow (they still use missing and there are a couple of failures related to reflection). You can continue to fix other things...

piever:
@quinnj did you happen to test the scaling of your `buildcolumns` implementation in case of unknown schema? If you have better performance we could try to port IndexedTables to use it, I suspect materializing the `NamedTuples` is bad for performance with several missable columns.

davidanthoff:
I wouldn’t expect unknown Schemas to be the problem, I think iterators that produce streams of values with heterogeneous types are the culprit (and those you easily get in a projection with ``Missing``, but not so easily with ``DataValue``). Those are really distinct.

JeffBezanson:
Got tests passing locally.

JeffBezanson:
Now with more TableTraits.

src/reduce.jl
```diff
@@ -431,23 +427,28 @@ end
 
 struct ApplyColwise{T}
     functions::T
-    names::Vector{Symbol}
+    names
```
JeffBezanson: Why this change?

quinnj:
Merge?

PR #185, opened 2018-08-13
WIP: new Tables.jl interface

quinnj:


quinnj:
Yes, this is based off a 0.7 update branch that Shashi started

src/tabletraits.jl
```diff
@@ -1,63 +1,13 @@
...
+sch(x::Type{T}) where {T <: Tuple} = NamedTuple{(Symbol("_$i") for i = 1:length(T.parameters)), T}
+Tables.schema(x::NextTable) = sch(eltype(x))
+Base.getproperty(x::Tuple, i::Int) = getfield(x, i)
+Tables.columns(x::NextTable) = columns(x)
+Tables.rows(x::NextTable) = rows(x)
 
-function TableTraits.getiterator(source::S) where {S <: NDSparse}
-    return rows(source)
-end
+table(x; kwargs...) = table(Tables.columntable(x); copy=false, kwargs...)
```
shashi: Maybe worth considering:

`Tables.columns(NamedTuple, x)` instead of a new `columntable`? `Tables.rows(Array, x)` would stand for `rowtable`. Mentioning the type seems to help explain what these methods do better than *table.

cc @andyferris

andyferris: Yeah, this is interesting. I'm not sure how I feel about `columntable` to be honest. Shashi's idea does seem to be extended, which is a big advantage.

davidanthoff:
As far as I can tell [IndexedTables.jl](https://github.com/JuliaComputing/IndexedTables.jl) is currently using ``DataValue`` for its missing value story. This PR here would now mean that one ends up with ``Missing`` values if one converts via this constructor here. I don't think that is good (and it would also be a breaking change). It would be especially confusing if one materialized a [Query.jl](https://github.com/queryverse/Query.jl) query into an ``IndexedTable``: the native missing value story in [Query.jl](https://github.com/queryverse/Query.jl) and [IndexedTables.jl](https://github.com/JuliaComputing/IndexedTables.jl) is ``DataValue``, but now when I combine them I all of a sudden get ``Missing``? That seems not ideal. Same for the whole Queryverse file IO story, that should really lead to tables that use the native missing value data type in [IndexedTables.jl](https://github.com/JuliaComputing/IndexedTables.jl), which right now seems to be ``DataValue``.

I think the way the [TableTraits.jl](https://github.com/queryverse/TableTraits.jl)/[IterableTables.jl](https://github.com/queryverse/IterableTables.jl) universe handles this is right: respect the native missing value story from each source and sink and when there is a conversion, always use the missing value story a given sink uses by default to represent missing values. So the approach had always been: if you convert something to a ``DataFrame``, use ``Missing``, but if you convert to something that uses ``DataValue``, use ``DataValue``.

If this package here at some later point migrates to ``Missing``, then that would be the time in my opinion to also adjust the constructor here accordingly.

davidanthoff:
Oh, and one other thing: I believe the old constructor handled iterators of ``Pair``s in some special way, didn't it? I didn't test this, but does this new version here preserve that?

davidanthoff:
> Oh, and one other thing: I believe the old constructor handled iterators of Pairs in some special way, didn't it? I didn't test this, but does this new version here preserve that?

And to answer my own question here: if one passed a stream of pairs (with named tuples as both the first and second element) to the existing constructor, it would "flatten" these ``Pair``s, and also make the columns from the first element from the pair the primary key columns automatically. I think that should be preserved, it is super useful.

piever:
I made a clumsy attempt to support this here: https://github.com/JuliaData/Tables.jl/pull/17 (trying to solve it in the general case: here we only want to unwrap pairs, i.e. an iterable of pairs should be converted to a pair of `Columns` object, of which the first `Columns` is used as primary keys). I'm not sure whether we should go for the general `unwrap` implementation or just do this for Pairs somehow.

shashi:
Looks like this can be closed now?

PR #194, opened 2018-11-16
Remove generated functions from getindex

joshday:
This simplifies `ith_all`, which is called by `getindex`.  I see a slight speedup with  
```julia
@btime include("test/runtests.jl")
```
after the change.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -31,15 +31,7 @@ right(x, y) = y
 
 # tuple and NamedTuple utilities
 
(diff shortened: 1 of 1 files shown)


shashi:
cool!

piever:
Maybe this is not the right place to ask. I'm really happy about this change but also a bit confused: how can the `map` version be as fast as the generated function? I've benchmarked and they seem comparable but I though the generated version should be faster. Is this due to some improvements in Julia 1.0?

PR #195, opened 2018-11-21
remove deprecations

joshday:
Removing deprecations that have been around for a year and many releases

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,9 @@ julia:
   - 0.7
   - 1.0
   - nightly
+matrix:
+  allow_failures:
+    - julia: nightly
 notifications:
   email: false
 # uncomment the following lines to override the default test script
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,35 @@
+environment:
+  matrix:
+  - julia_version: 0.7
+  - julia_version: 1
+  - julia_version: nightly
+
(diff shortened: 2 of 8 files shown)


piever:
Could you also add `IndexedTable` as a synonym for `NextTable` (I think that was the original plan)? Or you think this should be done in a future PR?

joshday:
Do you have a link to a conversation somewhere?  The idea is to rename `NextTable` to `IndexedTable`?

One of the lines this PR removes is 
```julia
Base.@deprecate_binding IndexedTable NDSparse
```

although `IndexedTable` isn't exported.

piever:
`IndexedTable` used to be what is now `NDSparse` and it was the only table type offered here. Then @shashi had the idea to create a table where some columns are sorted and to call this `IndexedTable` instead. However, not to subtly change meaning of the word, `IndexedTable` was deprecated and the thing was called `NextTable` instead. Now it's probably a safe time to bring `IndexedTable` back to the intended meaning. The design of the API is discussed in https://github.com/JuliaComputing/JuliaDB.jl/issues/80 and implemented i nhttps://github.com/JuliaComputing/IndexedTables.jl/pull/85

joshday:
Thanks for the links!  Since this is ready to go I'm going to merge it, but that will be next on my list of things to do.

PR #197, opened 2018-11-28
Replace: OnlineStats -> OnlineStatsBase, Change name: NextTable -> IndexedTable

joshday:
- Replace OnlineStats dependency with OnlineStatsBase
- Change `NextTable` -> `IndexedTable` (with deprecation warning for `NextTable`)


quinnj:
Great stuff here :)

PR #199, opened 2018-12-03
Tables Integration

joshday:
This PR:

- Adds `tables.jl` which provides the Tables integration
- renames `table.jl` -> `indexedtable.jl` to avoid confusion

My approach is to define the Tables interface for `Columns`, which provides the basis for integrating `IndexedTable` and `NDSparse`.

@quinnj : I haven't been following Tables development closely enough.  Is this sufficient to remove the TableTraits/TableTraitsUtils/IteratorInterfaceExtensions dependencies?

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,8 +10,28 @@ be used on its own for efficient in-memory data processing and analytics.
 
 ## Data Structures 
 
-- **The two table types in IndexedTables differ in how data is accessed.**
-- **There is no performance difference between table types for operations such as selecting, filtering, and map/reduce.**
+IndexedTables offers two data structures: `IndexedTable` and `NDSparse`.
+
+- **Both types store data _in columns_**.
+- **`IndexedTable` and `NDSparse` differ mainly in how data is accessed.**
+- **Both types have equal performance fr Table operations (`select`, `filter`, etc.).** 
+
+
+## Quickstart
+
+```
+using Pkg
+Pkg.add("IndexedTables")
+using IndexedTables
+
+t = table((x = 1:100, y = randn(100)))
+
+select(t, :x)
+
+filter(row -> row.y > 0, t)
+```
+
+## `IndexedTable` vs. `NDSparse`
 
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -10,8 +10,28 @@ be used on its own for efficient in-memory data processing and analytics.
 
 ## Data Structures 
 
-- **The two table types in IndexedTables differ in how data is accessed.**
-- **There is no performance difference between table types for operations such as selecting, filtering, and map/reduce.**
+IndexedTables offers two data structures: `IndexedTable` and `NDSparse`.
+
+- **Both types store data _in columns_**.
+- **`IndexedTable` and `NDSparse` differ mainly in how data is accessed.**
+- **Both types have equal performance fr Table operations (`select`, `filter`, etc.).**
```
quinnj: `for`

src/tables.jl
```diff
@@ -0,0 +1,26 @@
+#-----------------------------------------------------------------------# Columns 
+const TableColumns = Columns{T} where {T<:NamedTuple}
+
+Tables.istable(::Type{<:TableColumns}) = true
+Tables.materializer(c::TableColumns) = Columns
```
quinnj: Hmmm, I think we need to define the right `Columns`, `table`, and `ndsparse` functions that call `Tables.rows(x)` on the input argument, like [here](https://github.com/JuliaComputing/IndexedTables.jl/pull/185/files#diff-17226d5590764ff7e308bfa5045dacfaR16).

quinnj:
This looks pretty good so far; I left a comment on needing the "catchall" constructor functions that call `Tables.rows(x)` on the inputs and then pass them to the typed constructors. I don't think we should remove the TableTraits yet, @davidanthoff and I are trying to schedule a discussion to hash out the details of https://github.com/JuliaData/Tables.jl/pull/45, which would allow any Tables.jl implementation to automatically get TableTraits support. I'll try to make some progress on that PR soon.

PR #204, opened 2019-01-07
WIP: Fix IterableTables integration, support both missing and DataValue

joshday:
This PR fixes the IterableTables integration and allows for optionally creating `Missing` or `DataValue` inside of `join`s.

Until some of the performance issues with missing can get sorted out, the plan is for IndexedTables/JuliaDB to support both `missing` and `DataValue`.  Since missing values are generated in only two places (loading data and joins), this isn't terribly difficult to achieve. 

I've added a `missingtype = Missing` keyword argument for `join`.  For consistency, in JuliaDB I plan to add a `missingtype = Missing` keyword argument for `loadtable` and `loadndsparse`.


piever:
Nice! We should probably also put back the specialization for `arrayof(::DataValue{T}) where {T}` to return `DataValueArray{T}` (or it could also return like `Columns{DataValue{T}, NamedTuple{(:hasvalue, :value), Tuple{arrayof(Bool), arrayof(T)}}}`, which would be similar to the methods for `Pair`). We may also need to add some tests for collection with datavalues (meaning, `collect_colums`). I think that needs something like:

```julia
const default_initializer = ArrayInitializer(t -> t<:Union{Tuple, NamedTuple, Pair, DataValue})
```

so that datavalues are stored efficiently as an array with the `value` field and another array with the `hasvalue` field (this requires StructArrays master to work at the moment, but I can tag at any time).

davidanthoff:
I'll try to review this tomorrow.

davidanthoff:
Actually, this is still WIP? Can you ping me when it is ready to be reviewed?

joshday:
Yep, sure thing @davidanthoff

joshday:
Okay @davidanthoff this should be good to go.  The only difference from the previous IterableTables integration code is this (first try IterableTables.jl interface, then try Tables.jl interface).

```julia
function table(iter; copy=false, kw...)
    if TableTraits.isiterable(iter)
        table(collect_columns(IteratorInterfaceExtensions.getiterator(iter)); copy=copy, kw...)
    elseif Tables.istable(typeof(iter))
        table(Tables.columntable(iter); copy=copy, kw...)
    else
        throw(ArgumentError("input satisfies neither IterableTables.jl nor Tables.jl"))
    end
end
```

src/utils.jl
```diff
@@ -1,3 +1,26 @@
...
+unwrap(x) = x
+unwrap(x::DataValue) = get(x)
+
+ismissingtype(T, ::Type{Missing}) = Missing <: T 
+ismissingtype(T, ::Type{DataValue}) = T <: DataValue
+
+_ismissing(x) = ismissing(x)
+_ismissing(x::DataValue) = isna(x)
+
+Base.similar(T::Type{DataValueArray{S}}, n::Int) where {S} = DataValueArray(Vector{S}(undef, n))
```
piever: Can we dispatch on `n::Dims` instead? Or potentially add two methods? The machinery in StructArrays needs the `similar(T, ::Dims)` method.

joshday: done.

davidanthoff:
I finally found time to look through this. I think there are still a number of things that aren't correct, but hopefully they are easy to fix.

The first is that the iterator of named tuples that is returned by ``IteratorInterfaceExtensions.getiterator`` must *always* use ``DataValue`` for missing values, even if a column in ``IndexedTable`` uses ``Vector{Union{Missing,T}}``. This obviously wasn't an issue in previous versions because they never had those kind of columns, but now we need to take care of that. https://github.com/JuliaComputing/IndexedTables.jl/pull/211 is one attempt to do so.

The next one is less clear, but I still wanted to bring it up. The traditional philosophy for TableTraits.jl had been that each sink picks its own way on how to represent missing values (Missing or DataValue or whatever), and then whatever source you convert into that sink via TableTraits.jl, you always end up with that choice of missing value representation (i.e. the choice of the *sink*), regardless of how your source represents missing value. Right now the implementation here will instead give you different missing value representations depending on the source you start with: say you convert a ``DataFrame`` to an ``IndexedTable``, then you end up with an ``IndexedTable`` with ``Missing`` columns. If you convert anything from the Quervyerse to an ``IndexedTable``, though, you end up with ``DataValue`` columns. I don't find that ideal: I think it would make more sense for ``IndexedTable`` to have a default missing value story (whatever it is), and then always pick that regardless of how the source of tabular data is representing missing values. It just seems odd that say someone loads a stat file via [StatFiles.jl](https://github.com/queryverse/StatFiles.jl), that you would get a different missing value story inside your ``IndexedTable`` than if you load a file via [CSV.jl](https://github.com/JuliaData/CSV.jl).

The final issue is what happens when one converts an ``IndexedTable`` to a ``DataFrame`` when the ``IndexedTable`` has a column with ``DataValue``s. When that conversion went through the TableTraits.jl machinery in previous versions, one ended up with a ``DataFrame`` with a ``Union{Missing,T}`` column. I think that is the correct semantics: ``DataFrame`` has made a clear choice what its missing value story is, and so I think it makes sense to respect that. With the current version, one instead ends up with ``DataValue`` columns in the ``DataFrame``, which in my mind is a regression. I *think* (but am not 100% sure) this is actually entirely unrelated to anything from Queryverse at this point, because I believe that conversion is now entirely handled by Tables.jl. I'm not sure, though, because I don't fully understand the Tables.jl code anymore. In any case, I think the end result is not good and we should probably try to get back the behavior that we had in previous versions.

PR #211, opened 2019-02-11
Fix TableTraits source implementation

davidanthoff:
TableTraits.jl sources *must* use ``DataValue`` in their row iterator to be compliant with the interface, even if the source uses ``Union{Missing,T}`` in its internal implementation. This PR adds that behavior.


joshday:
Cool, thanks for taking this on!

Issue #103, opened 2017-12-03
Allow a default value?


shashi:
Missing indices now return a `KeyError`. Let's track the arguments for and against allowing a default value (such as 0) in NDSparse, so that it can behave something like an Array.

Pros:
- More consistency with arrays

Cons:
- Indices then are essentially unbounded (is this a problem in practice?)

Unclear:
- What should join operations do? `innerjoin` and `outerjoin` become equivalent since keys can be found in either table.

aaowens:
I know this is an old issue but I just had need for this feature in my project. I'm looking up a key which may not exist. If it doesn't exist, I would like to obtain `missing`.

I tried to just implement this manually, but it doesn't seem like `haskey` is implemented for NDSparse. Was this intentional? I could submit a PR for it.

joshday:
We should have a `haskey` method.  A PR would be great!  

I lean towards `get` and `get!` methods over default values.  Those should be straightforward to implement as well if you're up for it.  Otherwise I'll add them when I can.

piever:
`haskey` is nice but I agree that `get` and `get!` methods sounds like the ideal solution (as well as `get` with a function argument to be run only if the value is not found, which we could even use to implement `getindex`, to avoid code duplication).

PR #215, opened 2019-03-04
Add haskey, get, tests

aaowens:
See https://github.com/JuliaComputing/IndexedTables.jl/issues/103 .

I didn't add `get!`. I'm not really sure what the appropriate value to add would be, since an element of ndsparse is a named tuple. If the default value is missing, it might make sense to add a named tuple with all values missing, but this wouldn't work in all cases.

### DIFF
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -11,6 +11,24 @@ function _getindex_scalar(t, idxs)
     t.data[first(i)]
 end
 
+Base.haskey(t::NDSparse, idxs) = (flush!(t); _haskey(t, idxs))
+
+_haskey(t::NDSparse{T,D}, idxs::D) where {T,D<:Tuple} = _haskey_scalar(t, idxs)
+
+function _haskey_scalar(t, idxs)
+    i = searchsorted(t.index, convertkey(t, idxs))
+    return length(i) == 1
+end
+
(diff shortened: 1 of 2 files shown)


src/indexing.jl
```diff
@@ -11,6 +11,24 @@ function _getindex_scalar(t, idxs)
     t.data[first(i)]
 end
 
+Base.haskey(t::NDSparse, idxs) = (flush!(t); _haskey(t, idxs))
```
joshday: Why split this into three functions that have the same method signature?

src/indexing.jl
```diff
@@ -11,6 +11,24 @@ function _getindex_scalar(t, idxs)
...
+
+_haskey(t::NDSparse{T,D}, idxs::D) where {T,D<:Tuple} = _haskey_scalar(t, idxs)
+
+function _haskey_scalar(t, idxs)
+    i = searchsorted(t.index, convertkey(t, idxs))
+    return length(i) == 1
+end
+
+Base.get(t::NDSparse, idxs, default) = (flush!(t); _get(t, idxs, default))
```
joshday: I'm also fond of @piever's suggestion to add a `get(function, t, idxs)` method if you're up for it.

aaowens: I added that and used it to implement `_getindex_scalar`.

shashi:
Thanks for this!! :+1:

PR #219, opened 2019-03-09
optimize sortpermby

piever:
This moves the `fast_sortable` logic inside `sortperm_by` and `sortperm_fast`: I followed the scheme of https://github.com/piever/StructArrays.jl/pull/48 (which hopefully can be closed once this is merged) on `fast_sortable`: `StringArray{String}` goes to `StringArray{WeakRefString}`, `StringArray` gets pooled and on pooled arrays we use the same logic as what was implemented in PooledArrays (it needs the companion PR https://github.com/JuliaComputing/PooledArrays.jl/pull/21 to make sure we never end up doing the `fast_sortable` computation on PooledArrays twice). 

I hope this can restore good performance on `groupby` (end, more importantly, allows further performance tweaking without needing to change StructArrays), but have not benchmarked.

cc: @shashi @joshday

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -5,6 +5,6 @@ WeakRefStrings 0.5.7
 IteratorInterfaceExtensions 0.1.0
 TableTraits
 Tables
-StructArrays 0.2.0
+StructArrays 0.2.3
 DataValues 0.4.6
 TableTraitsUtils
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.7
 OnlineStatsBase 0.9.1
-PooledArrays 0.4.1
+PooledArrays 0.5.1
 WeakRefStrings 0.5.7
 IteratorInterfaceExtensions 0.1.0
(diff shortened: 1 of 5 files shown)


piever:
Actually, as both `sortperm` and `refine_perm!` are already optimized for PooledArrays, here I only need to pool strings before calling a sorting function, so I've renamed the preprocessing function `poolstrings`.

piever:
I wonder whether we actually need the intermediate `convert(StringArray{WeakRefString}, x)` step for string arrays, shouldn't it be enough to pool those?

piever:
I've improved `sortperm` and `refine_perm!` on PooledArrays in StructArrays (to avoind comparing strings but comparing the `UInt` value instead), so to optimize here it's enough to `pool` before running the sorting functions. For that it relies on https://github.com/piever/StructArrays.jl/blob/master/src/sort.jl#L12 so tests will pass here as soon as the StructArrays release is tagged and I can bump the dependency here.

Issue #165, opened 2018-04-23
groupby with pair selectors: unnecessary computation


piever:
```julia
julia> using IndexedTables

julia> name=["John", "Sally", "Kirk"];

julia> age=[23., 42., 59.];

julia> children=[3,5,2];

julia> t = table(name, age, children, names = [:name, :age, :children]);

julia> f(x) = (println(x); 1)
f (generic function with 1 method)

julia> groupby(identity, t, (:children => f,), flatten = true)
3
5
2
3
5
2
Table with 3 rows, 3 columns:
children  name     age
───────────────────────
1         "John"   23.0
1         "Sally"  42.0
1         "Kirk"   59.0
```

I think this can/should be optimized to compute `f` only once per row.

PR #222, opened 2019-03-17
reuse pooling from sortpermby in grouping algorithms

piever:
This has two advantages:

1. If data needs to be pooled, we only pool once and use the pooled data on `roweq`, which is much more efficient.
2.If the selection in `by` needs to compute something (i.e. `name => col => func`), we only compute it once (issue #165).

### DIFF
--- a/src/reduce.jl
+++ b/src/reduce.jl
@@ -121,7 +121,7 @@ function groupreduce(f, t::Dataset, by=pkeynames(t);
     if !isa(by, Tuple)
         by=(by,)
     end
-    perm, key = sortpermby(t, by, cache=cache, return_rows=true)
+    perm, key = sortpermby(t, by, cache=cache, return_keys=true)
 
     fs, input, T = init_inputs(f, data, reduced_type, false)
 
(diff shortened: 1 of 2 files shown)


src/reduce.jl
```diff
@@ -215,17 +214,16 @@ function groupby(f, t::Dataset, by=pkeynames(t);
         by = (by,)
     end
 
-    key = by == () ? fill((), length(t)) : rows(t, by)
-
     fs, input, S = init_inputs(f, data, reduced_type, true)
 
     if by == ()
+        key = fill((), length(t))
```
shashi: Why this line?

src/sortperm.jl
```diff
@@ -23,7 +23,7 @@ function best_perm_estimate(perms, cols)
     return (bestmatch, bestperm)
 end
 
-function sortpermby(t, by; cache=false)
+function sortpermby(t, by; cache=false, return_rows=false)
```
shashi: Name suggestion: `return_sorted`?

src/reduce.jl
piever: ups, well-spotted!

src/sortperm.jl
piever: That may be confusing though because the rows are not actually sorted (they are sorted according to `perm`), maybe `return_keys`? We end up calling that thing `keys` very often.

PR #223, opened 2019-03-18
Simplify grouping iteration mechanism

piever:
This uses the new `lazygroupmap` (bikeshedding on the name wanted) which is a low-level function to abstract the common mechanism in the grouping functions `groupby` and `groupreduce`. `lazygroupmap(f, keys, perm)`  returns an iterator that applies `f` to `(key, perm, idxs)` where `key` is the key of the group (as a struct), `perm` is the permutation and `idxs` is the range in `perm` that corresponds to the value `key`. Both `groupby` and `groupreduce` are special cases of this operation. `lazygroupmap` (implemented in https://github.com/piever/StructArrays.jl/pull/57/files, so tests will fail until that is merged and tagged) also takes care of optimizing comparison (to compute `idxs` only the refs of pooled data are compared) so it should have good performance.

EDIT: renamed `lazygroupmap` to `maptiedindices` (`tiedindices` is the iterator that iterates the key and range in `perm` corresponding to that value, so I figured `maptiedindices` is the natural name for this).

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -5,6 +5,6 @@ WeakRefStrings 0.5.7
 IteratorInterfaceExtensions 0.1.0
 TableTraits
 Tables
-StructArrays 0.2.3
+StructArrays 0.3.0
 DataValues 0.4.6
 TableTraitsUtils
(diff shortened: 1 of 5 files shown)


piever:
This passes tests now that the StructArrays version is tagged. Will merge in a couple of days if there are no objections.

Issue #151, opened 2018-04-03
Iterator roadmap


piever:
List of things to do concerning the transition to a more `Iterator` based style.

Getting rid of inference:

- [ ] port `join` to `collect_columns`

- [ ]  port `broadcast` to `collect_columns`

- [ ] remove other uses of `_promote_op` that I'm missing

Iterator compatibility / better composability:

- [ ] Add method `ndsparse(x)` where `x` is an iterator that iterates pairs

- [ ] Adapt `flatten` to also work when fed an iterator (so that when calling `groupby` with `flatten=true` we can skip one `collect_columns` step

- [ ] Add option `lazy=true` to `groupby`, `groupreduce`, etc... to return the iterator rather than the collected version

- [ ] See which functions can take an iterator as input. Decide what to do for those that cannot (for example `groupby` can't really be done on an iterator as we need the result of `sortperm`, but we could maybe add a trait to iterators to see if `sortperm` is known somehow, some sort of `HasSortPerm()`)

A concern is type piracy. For example, I would like to filter lazily and `reduce` in one go, but we can't add a `reduce` method for a general iterator as that's type piracy, so maybe we should define our own `AbstractRowIterator` for dispatch purposes. Then all lazy algorithms would return a subtype of `AbstractRowIterator`. Then maybe `RowIterator(t, select)` would iterate on `Rows(t, select)`.

We also need to think hard as to what this would mean for the distributed case. I don't have a strong intuition about that.

PR #225, opened 2019-03-22
implement join as iterator and remove reliance on inference

piever:
This adds an iterator that, given `lkeys, rkeys, lperm, rperm` iterates on the joint ranges `(l0:l1, r0:r1)` that correspond to the sorted keys (it uses some mergesort like technique to combine the two columns). The plan is to convert `groupjoin` (and all the other joins) to an iterator based version that uses this. Requires StructArrays master.

It is the last challenging bit of #151 to remove our dependency on type inference.


piever:
In retrospect, I suspect `GroupJoinPerm` may even live in StructArrays (as a general companion too `GroupPerm`, so that StructArrays is already good enough for basic data manipulations) and here I would just use it to reimplement the user-friendly versions (`join`, `groupjoin`, `antijoin`, etc...).

piever:
This simplifies the code greatly and passes tests (at least locally), with two sacrifices:

- performance (which hopefully can be fixed)
- keepkeys option, which I didn't understand and seems a bit awkwardly implemented (with `@label` and `@goto`). 

I can work on performance and benchmark carefully, but wanted to make sure whether it is OK to remove keepkeys?

src/join.jl
```diff
@@ -1,182 +1,15 @@
...
-                append!(data, map(x->init_group(), j:rr))
-            else
-                append!(lnull_idx, (1:length(j:rr)) .+ length(data))
-                _append!(Val{:right}(), f, data, lout, rout,
-                       ldata, rdata, 0, rperm[j:rr], lnull, rnull)
-            end
-        end
-    end
-    lnull_idx, rnull_idx
-end
```
shashi: what have you done with my favorite secret function?

on a serious note this is awesome heroic work as always!

piever: I remember you mentioning it was the ugliest function you ever wrote :) I may need to bring some of that back to optimize the `f === concat_tup` case though.

piever:
Even if not automated, do we have some `join` benchmarks I can try to make sure I don't cause a performance regression here?

piever:
From quick benchmarking it seems that `groupjoin` is actually already competitive with the old version, whereas regular `join` is slower so we may need to optimize `collect_columns_flattened` somehow.

piever:
As an overall strategy I would suggest to merge this soonish (after I test the "large data table" case and implement a specialization on `typeof(concat_tup)` as we had before for that) and maybe wait longer before releasing to see if some issues pop up.

shashi:
that is impressive, good job on passing CI!

I'll look to see if the allocations can be helped right now.

shashi:
are you sure `N=10_1000` in your demos?

src/join.jl
```diff
@@ -181,92 +10,68 @@ end
...
-        else
-            out_type = _promote_op(f, left_type, right_type)
-        end
-        if init_group === nothing
-            init_group = () -> similar(arrayof(out_type), 0)
-        end
-        if accumulate === nothing
-            accumulate = push!
+            Base.foreach(Iterators.product(riter, liter)) do (r, l)
+                push!(init[1], key)
```
piever: This is the optimized version in case it's a plain join with `f===concat_tup`

piever:
I've added back the optimization in case of plain join with `concat_tup` where it basically just pushes line by line as in the previous version, which helps cut down memory usage and is important when the rows are large and non isbits and actually doing `concat_tup` would be expansive (if we have better ideas on how to optimize that case, it's easy to add them in those lines).

I think this is good to go.

joshday:
So I just found out that `keepkeys` was necessary for keeping around the NDSparse keys that are not being joined on.  I don't think there's an alternative way for doing that.

joshday:
Example: how do I keep column `z` in the joined table?

```julia
julia> t = ndsparse((x=1:10,y=1:10), rand(10));

julia> t2 = ndsparse((x=1:10,z=1:10), rand(10));

julia> join(t, t2)
2-d NDSparse with 10 values (2-tuples):
x   y  │ 3          4
───────┼────────────────────
1   1  │ 0.0275186  0.62887
2   2  │ 0.873126   0.816695
3   3  │ 0.134846   0.393059
4   4  │ 0.453348   0.78895
5   5  │ 0.748839   0.189154
6   6  │ 0.878384   0.773631
7   7  │ 0.107017   0.711591
8   8  │ 0.817191   0.871991
9   9  │ 0.195588   0.3514
10  10 │ 0.870386   0.805576
```

piever:
I'm afraid that in the general case (meaning also `outer` option is allowed), this is a mess: for example keys would have to become nullable as sometimes I'm taking values that exist only in left but would need to fill in some of the keys that are only from right (and indeed in the previous code `keepkeys` felt a bit hacky with its `@label` and `@goto`). In my mind currently in `join` the keys should be the same in both tables and this thing that you are suggesting is more of a job for broadcast (at least, broadcast docs explicitly suggest that keys on one side are a subset of keys on the other side). I suspect that the thing you are looking for is this TODO https://github.com/JuliaComputing/IndexedTables.jl/blob/master/src/join.jl#L583.

If you agree with this solution I can try to implement the TODO for broadcast.

cwiese:
This may be late in the game.  We are trying to upgrade a huge application that has been in production for a couple years.  We rely on "keepkeys" .. just like SQL we need ability to keep or select keys in result set.    This is important even in 1:M join.   We are going to test broadcast or some other work around - however, we do use "keepkeys" in many place and are stuck on Julia 1.2 without a workaround

cwiese:
using rselect does not help - as the columns become data and not keys - and we have 2 million row tables ..noow neededing sorting to move to keys

`julia> t2 = ndsparse((1:10,1:10,1:10), rand(10))
3-d NDSparse with 10 values (Float64):
1   2   3  │
───────────┼─────────
1   1   1  │ 0.110031
2   2   2  │ 0.138255
3   3   3  │ 0.721263
4   4   4  │ 0.418394
5   5   5  │ 0.863919
6   6   6  │ 0.68232
7   7   7  │ 0.706424
8   8   8  │ 0.641431
9   9   9  │ 0.123797
10  10  10 │ 0.47846

julia> join(t1, t2; lkey=(1,), rkey=(1,), rselect=(1,2,3,), lselect=(1,2,3,) )
1-d NDSparse with 10 values (6-tuples):
1  │ 2   3   4   5   6   7
───┼───────────────────────
1  │ 1   1   1   1   1   1
2  │ 2   2   2   2   2   2
3  │ 3   3   3   3   3   3
4  │ 4   4   4   4   4   4
5  │ 5   5   5   5   5   5
6  │ 6   6   6   6   6   6
7  │ 7   7   7   7   7   7
8  │ 8   8   8   8   8   8
9  │ 9   9   9   9   9   9
10 │ 10  10  10  10  10  10

julia> join(t1, t2; lkey=(1,), rkey=(1,), rselect=(1,2,3,4,), lselect=(1,2,3,4,) )
1-d NDSparse with 10 values (8-tuples):
1  │ 2   3   4   5          6   7   8   9
───┼────────────────────────────────────────────
1  │ 1   1   1   0.532105   1   1   1   0.110031
2  │ 2   2   2   0.765344   2   2   2   0.138255
3  │ 3   3   3   0.415088   3   3   3   0.721263
4  │ 4   4   4   0.810639   4   4   4   0.418394
5  │ 5   5   5   0.0198196  5   5   5   0.863919
6  │ 6   6   6   0.216057   6   6   6   0.68232
7  │ 7   7   7   0.621441   7   7   7   0.706424
8  │ 8   8   8   0.370124   8   8   8   0.641431
9  │ 9   9   9   0.298708   9   9   9   0.123797
10 │ 10  10  10  0.769241   10  10  10  0.47846

PR #230, opened 2019-03-29
remove `_promote_op`

piever:
After the release of JuliaDB gets merged I think we might as well remove `_promote_op`: it's a private function that we no longer need.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -169,16 +169,6 @@ Base.@pure function strip_unionall(T)
     end
 end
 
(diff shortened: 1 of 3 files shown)


shashi:
So should we release IndexedTables, JuliaDB and then merge this and release again? Feel free to do so! :)

piever:
I've actually already tagged JuliaDB as the changes there didn't require the changes here to work, so I think we can just merge here and tag (I plan to do that tomorrow).

PR #233, opened 2019-04-04
Port remaining function to iteration framework

piever:
Now `flatten` can handle type change in the eltype of the iterator (and we can flatten a column of iterators) and aggregate only relies on the high-level `GroupPerm` and not on `roweq` (which I'd like to keep private to StructArrays).


piever:
On a separate note and for a separate PR, I don't really like the idea that `flatten` should only flatten on arrays of iterables. I think we should just error there like `Iterators.flatten`:

```julia
julia> Iterators.flatten([:a, :b]) |> collect
ERROR: MethodError: no method matching iterate(::Symbol)
```

piever:
I've changed it to actually test that all elements are iterable (rather than checking the eltype, as the eltype could be loose): I'm using a generated function so that in the type stable case this is still a no-op. Had to remove Julia 0.7 support (as things there were fixed by a hacky try catch), hope that's OK!

shashi:
that's ok

PR #236, opened 2019-04-10
use permute!! or copyto! instead of fastpermute!

piever:
Again, part of the plan to decouple IndexedTables from StructArrays internals. I'm also adding some small optimizations to avoid unnecessary allocations and to have things efficient also in the `Union{String, Missing}` case.

### DIFF
--- a/src/IndexedTables.jl
+++ b/src/IndexedTables.jl
@@ -7,7 +7,7 @@ using OnlineStatsBase: OnlineStat, fit!
 using StructArrays: StructVector, StructArray, foreachfield, fieldarrays,
     collect_structarray, staticschema, ArrayInitializer, refine_perm!, collect_structarray,
     collect_empty_structarray, grow_to_structarray!, collect_to_structarray!, replace_storage,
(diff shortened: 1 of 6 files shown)


joshday:
Dumb question: what is `Base.permute!!`?  There's no docs for it that I can find.

piever:
> Dumb question: what is Base.permute!!? There's no docs for it that I can find.

Not sure why it's not documented. It's the same as `permute!` but it also consumes the permutation vector:

```julia
julia> a = ["a", "b"];

julia> p = [2, 1];

julia> Base.permute!!(a, p);

julia> a
2-element Array{String,1}:
 "b"
 "a"

julia> p
2-element Array{Int64,1}:
 0
 0
```

I think we should use this instead of `permute!` in all cases where we are discarding the permutation vector afterwards to avoid extra allocations (`permute!` simply makes a copy of the permutation vector and then calls `permute!!`)

joshday:
Okay got ya, sounds good to me 👍

piever:
Can I tag a new point release so that I'll be free to remove this things from StructArrays eventually?

joshday:
Absolutely!

Issue #239, opened 2019-04-15
Column manipulation API cleanup


piever:
Discussing the DataFrames API over at https://github.com/JuliaData/DataFrames.jl/issues/1695#issuecomment-483163054, I got the impression that our column manipulation developed organically over time and may need a little bit of clean-up.

Specifically:

**Problem**: `pushcol` and `setcol` seem largely overlapping (if the name is new in `setcol` the column gets added). The method `setcol(table, name, col)` and `setcol(table, name => col)` are redundant and the multi argument metod `setcol(table, pair1, pair2, pair3)` can be a bit confusing combined with `setcol(table, name, col)`. The name is a bit puzzling as `setcol` is singular but can be applied to many columns: it seems like the preference in Julia in this case is to go with the plural (say `sum(v, dims = i)`). Also, it's inconsistent that the getter / setter for columns are `column` / `setcol`.

**Proposed solution**: Deprecate both in favor of `mutate(table, changes::Pair...)`. This is consistent with the `replace(v, old, new)` to `replace(v, old => new)` deprecation in Julia Base and with DataFrames intended `mutate!` syntax. Here the corresponding JuliaDBMeta macro, called `@transform`, should probably become `@mutate`. Related to this, we could also consider adding a `mergecols` to merge columns of two tables.

**Problem**: `popcol` is a poor name for various reasons. Unlike `pop!`, it does not return the item it takes out of the collection, the singular is odd as it allows to remove many columns, unlike `pop!` one can specify which column to remove.

**Proposed solution**: rename to `deletecols` (consistent with DataFrames: https://github.com/JuliaData/DataFrames.jl/pull/1772), stop defaulting to last column. `deletecols(t, ncols(t))` is not that bad and it's not clear to me why it should be more common to delete the last column instead of any other. If we want special syntax for that we may add a special selector `Last()` (along the lines of `Keys()`) but I don't think that's necessary.

`renamecol` also has similar issues as above and I think the correct signature should be `renamecols(table, changes:::Pair...)` (DataFrames uses `rename`, but I think it's nice to be consistent and use `...cols` for everything).

With `insertcol` and friends I would also rename to `insertcols` and friends (similar to `DataFrames.insertcols!`) and again only support the `name => col` interface (like DataFrames does). For example `insertcols(table, ind, name => v)` or `insertcolsbefore(table, col, name => v)`.

cc: @bkamins and @nalimilan

nalimilan:
[...]

PR #240, opened 2019-04-15
 deprecate popcol, pushcol and setcol

piever:
First part of #239. Here I'm deprecating `popcol(t, cols)` in favor of `select(t, Not(cols))` and `pushcol` and `setcol` are replaced by `transform`. I changes some generated code into normal code for readability as it is only needed for `rename!` now. I also did a bit of clean-up removing `ColDict` methods that were no longer needed.

### DIFF
--- a/src/IndexedTables.jl
+++ b/src/IndexedTables.jl
@@ -33,9 +33,9 @@ export
     column, columns, convertdim, dimlabels, flatten, flush!, groupby, groupjoin,
     groupreduce, innerjoin, insertafter!, insertbefore!, insertcol, insertcolafter,
     insertcolbefore, leftgroupjoin, leftjoin, map_rows, naturalgroupjoin, naturaljoin,
-    ncols, ndsparse, outergroupjoin, outerjoin, pkeynames, pkeys, popcol, pushcol,
-    reducedim_vec, reindex, renamecol, rows, select, selectkeys, selectvalues, setcol,
-    stack, summarize, table, unstack, update!, where, dropmissing, dropna
+    ncols, ndsparse, outergroupjoin, outerjoin, pkeynames, pkeys,
+    reducedim_vec, reindex, renamecol, rows, select, selectkeys, selectvalues,
+    stack, summarize, table, transform, unstack, update!, where, dropmissing, dropna
 
 const Tup = Union{Tuple,NamedTuple}
 const DimName = Union{Int,Symbol}
--- a/src/columns.jl
+++ b/src/columns.jl
@@ -530,6 +530,7 @@ transform(t, args...) = @cols transform!(t, args...)
 @deprecate pushcol(t, args) transform(t, args)
 
 @deprecate popcol(t, args...) select(t, Not(args...))
+@deprecate popcol(t) select(t, Not(ncols(t)))
 
 """
     insertcol(t, position::Integer, name, x)
--- a/src/columns.jl
+++ b/src/columns.jl
@@ -392,7 +392,7 @@ end
 Base.getindex(d::ColDict, key) = rows(d[], key)
 Base.getindex(d::ColDict, key::AbstractArray) = key
 
-function Base.setindex!(d::ColDict, x, key::Union{Symbol, Int})
+function Base.setindex!(d::ColDict, x, key::Union{Symbol, Integer})
     k = _colindex(d.names, key, 0)
     col = d[x]
     if k == 0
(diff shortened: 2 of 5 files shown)


piever:
One issue that came up is that `OnlineStats` also exports `transform`, should we import it from there and define things only for `AbstractIndexedTable` (I'm not sure if these methods also apply to `AbstractNDSparse`)?

nalimilan:
Maybe let's define `transform` in a common API package before you make a release? That function is already defined in Tables.jl, but not exported and with a lazy approach. Not sure whether we want to use that, or define it in another package.

piever:
Definitely. I'm not sure I like the idea of keeping this things in Tables though (I'm getting convinced that it'd be better to try and "slim Tables" as much as possible to some minimal interface along the lines of TableTraits, plus having an operation that's lazy or eager depending on input type doesn't feel ideal). 

A possible plan could be as follows:

- Try to keep Tables.jl minimal (I personally think it should aim to be as lightweight as say TableTraits)
- Move lazy table operations like `select` and `transform` to a new TableQueries.jl (and try to move all things not strictly needed in Tables to define the interface there)
- Use the existing TableOperations for eager data manipulation functions

Does that make sense? We could start by repurposing TableOperations (it already exists and has no longer anything relevant as its content was moved into Tables) so that we can extend `transform` and `select` from there.

nalimilan:
I merely suggested moving empty definitions to some package. AFAIK TableOperations provides actual implementations, not just stubs. Cc: @quinnj, @bkamins

quinnj:
I'm worried about having a common definition for `transform`, which is why I didn't export it from Tables.jl. Where have we landed on that? I know there's been a flurry of discussion in DataFrames/here. Slimming down Tables.jl is very high on my list (hoping to get to it this week), but I'm not exactly sure what makes the most sense in terms of breaking things up. I've kind of been thinking that we'd have actual Tables.jl-specific code in TableOperations.jl, so I kind of like the idea of having TableQuery.jl to house the basic, empty interface functions that everyone can extend.

bkamins:
My understanding of discussions on Slack/Discourse is that:
* we need a slim package exporting common names
* it should mostly contain empty definitions; packages can differ how they actually implement the methods
* one key issue, related with Julia multiple dispatch logic is that it would be ideal that we agree on a common placement of key distinguishing positional parameter; a natural choice is that the first parameter in these functions should be table type (this is important because `fun(x, df::TableType1)` and `fun(df::TableType2, x)` will be ambiguous otherwise;
* regarding `transform`, while I like the name it now occurs to me that it might be too common (e.g. many other unrelated packages might use the same name), maybe `mutate` is better because it is less likely to be used (not sure here).

nalimilan:
I don't really have an opinion about where the methods should live. A separate package is OK. Maybe TableFunctions?

Regarding potential conflicts with `transform`, it would be too bad not to use the most natural names just because of that. At least, defining them in a very lightweight package will be better than claiming them from a specific package that others cannot depend on. Hopefully we'll eventually find a solution to extend functions without loading packages that define them.

bkamins:
> Regarding potential conflicts with `transform`

This is a mild issue (as I have said - I was not really sure what is best). More important is to have an agreement on the location of the "key" argument allowing to safely dispatch the methods (but I think it should not be hard to agree on).

piever:
> Slimming down Tables.jl is very high on my list (hoping to get to it this week), but I'm not exactly sure what makes the most sense in terms of breaking things up. I've kind of been thinking that we'd have actual Tables.jl-specific code in TableOperations.jl, so I kind of like the idea of having TableQuery.jl to house the basic, empty interface functions that everyone can extend.

In my mind there are three distinct things (essentially the same as what you are proposing):
1. basic interface for tables (`rows`, `columns`, rowaccess`, columnaccess`)
2.things that are useful for tables (e.g. basic lazy queries like `select` and `transform`)
3. placeholder for generally useful table operations.

Even though TableOperations is more 2. than 3. at the moment, I would actually propose to repurpose TableOperations for 3 (this is breaking but I don't think anyone is really using TableOperations as all its functionality lives in Tables) and move `Tables.select` and `Tables.transform` to a new TableQuery. The rationale is that a TableQuery would have composable lazy functions (i.e. queries) whereas TableOperations would have actual eager operations that return a collected table.

How to split the current Tables package between 1. and 2. is not completely obvious but in my mind orthogonal to where to put the `select` and `transform` placeholders, so we should probably discuss it in a separate thread.

PR #241, opened 2019-04-15
Switch renamecol to rename and insertcol to insertcols, add multiple column version

piever:
This is the last part of #239, renaming `renamecol` to `renamecols` and deprecating `insertcol` to `insertcols` plus multiple columns version.

Two doubts:

1. Should the name be `renamecols` or just `rename`? After all we are obviously renaming columns (it wouldn't make to rename rows). Maybe we should just be consistent with `DataFrames` and call it `rename`.

2. I've interpreted the "multiple argument" version of `insertcols!` to behave as follows:

```julia
julia> t = table(rand(3), rand(3), rand(3), names = [:a,:b,:c])
Table with 3 rows, 3 columns:
a         b         c
────────────────────────────
0.509087  0.763431  0.557554
0.217274  0.801138  0.160908
0.462436  0.506603  0.921267

julia> insertcols(t, 2, :d => rand(3), :e => rand(3))
Table with 3 rows, 5 columns:
a         d         e         b         c
────────────────────────────────────────────────
0.509087  0.212275  0.683083  0.763431  0.557554
0.217274  0.49579   0.516921  0.801138  0.160908
0.462436  0.404024  0.982535  0.506603  0.921267

julia> insertcolsbefore(t, :c, :d => rand(3), :e => rand(3))
Table with 3 rows, 5 columns:
a         b         d          e         c
─────────────────────────────────────────────────
0.509087  0.763431  0.730766   0.263156  0.557554
0.217274  0.801138  0.844378   0.915973  0.160908
0.462436  0.506603  0.0846812  0.163381  0.921267
```

That is, we pass a unique index and all columns are inserted starting from there. I thought the version with multiple indices would be too confusing as the indices would change position when the first columns are inserted. Does this make sense?

### DIFF
--- a/src/IndexedTables.jl
+++ b/src/IndexedTables.jl
@@ -34,7 +34,7 @@ export
     groupreduce, innerjoin, insertafter!, insertbefore!, insertcols, insertcolsafter,
     insertcolsbefore, leftgroupjoin, leftjoin, map_rows, naturalgroupjoin, naturaljoin,
     ncols, ndsparse, outergroupjoin, outerjoin, pkeynames, pkeys,
-    reducedim_vec, reindex, renamecols, rows, select, selectkeys, selectvalues,
+    reducedim_vec, reindex, rename, rows, select, selectkeys, selectvalues,
     stack, summarize, table, transform, unstack, update!, where, dropmissing, dropna
 
 const Tup = Union{Tuple,NamedTuple}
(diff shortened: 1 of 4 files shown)


joshday:
1.  I vote for `rename`.
2. Does it make sense? Yes.

piever:
OK, I'm going with `rename` (simpler and in agreement with DataFrames).

PR #243, opened 2019-04-17
Update reduce.jl

yakir12:
Solves https://github.com/JuliaComputing/JuliaDB.jl/issues/278
Added documentation to the groupby(t, by; usekey = true)


piever:
Thanks!

PR #245, opened 2019-04-23
use IndexedTables rather than StructArrays utils

piever:
There already is a `foreach` in IndexedTables that can do the same job as `foreachfield` from StructArrays so it's probably safer to not rely on external "utility" functions.


piever:
Better to wait a bit before merging here as `foreachfield` is actually so useful for `StructArrays` that I may document it / export it and make it part of the official API, so maybe this change is not needed.

PR #250, opened 2019-05-07
Simplify table fallback constructor logic

quinnj:


src/tabletraits.jl
```diff
@@ -46,11 +47,9 @@ function table(rows::AbstractArray{T}; copy=false, kwargs...) where {T<:Union{Tu
...
 
 function table(iter; copy=false, kw...)
-    if TableTraits.isiterable(iter)
-        table(collect_columns(IteratorInterfaceExtensions.getiterator(iter)); copy=copy, kw...)
-    elseif Tables.istable(typeof(iter))
+    try
         table(Tables.columntable(iter); copy=copy, kw...)
-    else
-        throw(ArgumentError("input satisfies neither IterableTables.jl nor Tables.jl"))
+    catch e
```
joshday: Should we include `e` in the error message to hint as to why `iter` can't be converted into a table?

quinnj: It will pretty much be the same message, so some packages have opted to remove the try-catch all together, though the error would be a Tables.jl error instead of an IndexedTables error.

joshday: I'd be okay with removing the try-catch.  Especially if that's what e.g. DataFrames does, so that behavior is somewhat consistent across packages.

quinnj: Well, DataFrames still has a try-catch, but TypedTables, CSV.write, SQLite all just let the error bubble up from Tables

joshday: Ah, okay.  Let's keep the try-catch then.  That's probably a better error to see for newcomers.

src/tabletraits.jl
```diff
@@ -46,11 +47,9 @@ function table(rows::AbstractArray{T}; copy=false, kwargs...) where {T<:Union{Tu
...
+        throw(ArgumentError("can't convert input to a table"))
```
davidanthoff: I think this is not a good solution: it entirely masks useful error messages that might be thrown in a source. For example, for all the queryverse file IO packages, one now gets this error message if one tries to open a non existing file, even though the underlying packages throw error messages that properly point out what the problem is.

quinnj:
While I agree in principle, I think this is a pretty poor example; checking `isfile` seems like pretty obvious argument validation before ever calling `table`

davidanthoff:
> checking `isfile` seems like pretty obvious argument validation before ever calling `table`

Not in a lazy loading scenario that the queryverse file IO packages implement: you can check whether a file exists during the `load` call, but whether that is the case or not is quite irrelevant. The actual file opening and loading only happens once the data is requested, and that is triggered from inside the `table` call, so that is when the error will be raised.

piever:
Should we just remove the `try ... catch` block? If `iter` is a complex iterator with some error inside, it will just error the first time it is collected with an unhelpful `"can't convert input to a table"`.

davidanthoff:
Yes, I think that would probably resolve this problem.

quinnj:
PR to remove try-catch #253

PR #253, opened 2019-05-26
Remove try catch from fallback table constructor to not mask errors

quinnj:
@joshday @davidanthoff


piever:
Thanks! I'll merge this tomorrow if there are no objections.

PR #254, opened 2019-06-25
Bump version and clean up dependency compat

quinnj:
cc: @joshday , I think this should be good

cc: @davidanthoff I cleaned up the compat versions a little here

also cc: @piever to take a look

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,5 +1,6 @@
 name = "IndexedTables"
 uuid = "6deec6e2-d858-57c5-ab9b-e6ca5bd20e43"
+version = "0.12.1"
 
 [deps]
 DataValues = "e7dc6d0d-1eca-5fa6-8ad6-5aecde8b7ea5"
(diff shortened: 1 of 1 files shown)


Project.toml
```diff
@@ -16,14 +17,14 @@ Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
...
-DataValues = "≥ 0.4.6"
-IteratorInterfaceExtensions = "^0.1, ^1"
-OnlineStatsBase = "≥ 0.9.1"
-PooledArrays = "≥ 0.4.1"
-StructArrays = "≥ 0.3.2"
-TableTraits = "^0.3, ^0.4, ^1"
-WeakRefStrings = "≥ 0.5.7"
-Tables = "≥ 0.2.4"
+DataValues = "0.4.6"
+IteratorInterfaceExtensions = "0.1.1, 1"
```
davidanthoff: I don’t think this needs to be bumped to 0.1.1, the previous 0.1 should be fine.

quinnj: Ah, I just copied what was in DataFrames

PR #263, opened 2019-08-08
Take special selectors from DataAPI

piever:
So that IndexedTables and DataFrames can use the same special selectors. I'm leaving a buffer time with a deprecation warning for `Not` as JuliaDB accepts a multiargument version, impossible without type pirating InvertedIndices. When we drop the deprecation we can start using the InvertedIndices struct.

cc: @bkamins


bkamins:
Looks good. Thank you!

PR #264, opened 2019-09-29
update transform and Not docstring

piever:
There was an outdated reference to `pushcol` in the `transform` docstring.


piever:
I'm taking the occasion to also update the `Not` docstring which suggests a deprecated functionality (`Not(args...)`, which we had to deprecate when we planned to use `Not(x)` from InvertedIndeces).

PR #273, opened 2020-02-12
Support Tables.jl 1.0

quinnj:
Will require https://github.com/JuliaArrays/StructArrays.jl/pull/116/ + release first.


piever:
Thanks! I'll wait a little bit before merging here, because to tag StructArrays I need to tag here first (latest StructArrays removes some internal functions on which unfortunately IndexedTables was relying).

PR #278, opened 2020-05-06
Use StatsBase.transform

joshday:


joshday:
Ref: https://github.com/JuliaData/JuliaDB.jl/issues/324

PR #281, opened 2020-07-28
Fix newname => selector pairs

### LINKED ISSUE #279: select Tuple{Selection} does not seem to work as per docs
Trying to use `select(table, (:Name => ::Selection,))` does not seem to work as per my interpretation of the docs.
From the documentation:

> Selection is a type union of many types that can select from a table. It can be:
> 
> 1. Integer – returns the column at this position.
> 2. Symbol – returns the column with this name.
> 3. Pair{Selection => Function} – selects and maps a function over the selection, returns the result.
> 4. AbstractArray – returns the array itself. This must be the same length as the table.
> 5. **Tuple of Selection – returns a table containing a column for every selector in the tuple. The tuple may also contain the type Pair{Symbol, Selection}, which the selection a name. The most useful form of this when introducing a new column.**
> 6. Regex – returns the columns with names that match the regular expression.
> 7. Type – returns columns with elements of the given type.

As per this, I would expect to be able to select columns under a new name using `select`.
E.g. If I call `select(table, (:Name => :OldName,))`, as this satisfies the type `Tuple{Pair{Symbol, Selection}}`.
Is this meant to work?

akdor1154:
This should fix #279 , that is, make `select(tbl, (:newname => :existing,))` work.

One note is that it's now not possible to to select `:col => mapfn` when `!(mapFn isa Function)`. I don't think it's possible to resolve this succinctly/with dispatch (see: https://discourse.julialang.org/t/functions-and-callable-methods/2983). A workaround for users using functors like this would be `select(tbl, (mapFn.(tbl.columns.col),))`.

If the above is not acceptable, then alternatively I could special-case the selector `Pair{Symbol, Union{Symbol, Int, String, AbstractArray}}`, but this seems less general that ideal.

### DIFF
--- a/src/columns.jl
+++ b/src/columns.jl
@@ -246,7 +246,6 @@ column(c, x) = columns(c)[colindex(c, x)]
 end
 
 column(t, a::AbstractArray) = a
-column(t, a::Pair{Symbol, <:AbstractArray}) = column(t, a[2])
 column(t, a::Pair{Symbol, <:Any}) = column(t, a[2]) # :newname => selector
 column(t, a::Pair{<:Any, <:Base.Callable}) = map(a[2], rows(t, a[1]))
 column(t, a::Pair{Symbol, <:Base.Callable}) = map(a[2], rows(t, a[1])) # need a tiebreaker for above two
--- a/src/columns.jl
+++ b/src/columns.jl
@@ -227,10 +227,10 @@ function _colindex(fnames::Union{Tuple, AbstractArray}, col, default=nothing)
         return 0
     elseif isa(col, Tuple)
         return 0
-    elseif isa(col, Pair{Symbol, <:Pair}) # recursive pairs
-        return _colindex(fnames, col[2])
-    elseif isa(col, Pair{<:Any, <:Any})
+    elseif isa(col, Pair{<:Any, <:Base.Callable}) # selector => mapfn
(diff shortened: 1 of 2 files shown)


src/columns.jl
```diff
@@ -227,10 +227,10 @@ function _colindex(fnames::Union{Tuple, AbstractArray}, col, default=nothing)
         return 0
     elseif isa(col, Tuple)
         return 0
-    elseif isa(col, Pair{Symbol, <:Pair}) # recursive pairs
-        return _colindex(fnames, col[2])
-    elseif isa(col, Pair{<:Any, <:Any})
+    elseif isa(col, Pair{<:Any, <:Function}) # selector => mapfn
```
joshday, line 230:
```suggestion
    elseif isa(col, Pair{<:Any, <:Base.Callable}) # selector => mapfn
```

joshday:
Thanks!  I'll blame GitHub's new notification system for missing this.

src/columns.jl
joshday: From just a cursory glance at this, we could use `Base.Callable` instead of `Function` to allow for callable types

akdor1154:
no worries haha. Added your change (thanks!), checked tests still pass, and squashed (left the overload commit separate on purpose for imo clearer history).

joshday:
Thanks again!

